PR #29, opened 2015-01-16
[bug 1115229] Simple snippet template for Hello

ckprice:
@Osmose r?

Link to sample: https://snippets.allizom.org/admin/base/snippet/77/

### DIFF
--- a/templates/hello/simple-snippet.html
+++ b/templates/hello/simple-snippet.html
@@ -1,0 +1,209 @@
+<!-- Any instances of {{ snippet_id }} are replaced with the ID number for the
+     snippet using this template. -->
+
+<div class="snippet" id="{{ snippet_id }}-hello-simple">
+    <a href="#" class="launch-hello">
+      <img class="icon" src="{{ icon }}" />
+    </a>
+    <p>{{ text|safe }}</p>
+
+    <script type="text/javascript">
+      //<![CDATA[
+
+      (function() {
+        var snippet = document.getElementById('{{ snippet_id }}-hello-simple');
+        snippet.addEventListener('show_snippet', function () {
+          var buttons = snippet.getElementsByClassName('launch-hello');
+
+          isTargetAvailable('loop', function loopAvailable () {
+            sendMetric('hello-eligible-view');
+            for (var i=0; i < buttons.length; i++) {
+              buttons[i].addEventListener('click', function (e) {
+                e.preventDefault();
+                Mozilla.UITour.showMenu('loop', function () {
+                  // bug 1113896
+                  document.addEventListener('click', function clickHidePanel (e) {
+                      Mozilla.UITour.hideMenu('loop');
+                      document.removeEventListener('click', clickHidePanel, false);
+                  }, false);                  
+                });
+
+                // bug 1115227
+                Mozilla.UITour.registerPageID('hello-tour_OpenPanel_snippet-{{ snippet_id }}');
+
+                // bug exists where teardown tabs will stop .observe
+                // this is to ensure it is re-started
+                document.addEventListener('visibilitychange', handleVisibilityChange);
+
+                sendMetric('hello-open');
+              }, false);
+            }
+          }, function loopNotAvailable () {
+            sendMetric('hello-ineligible-view');
+            for (var i=0; i < buttons.length; i++) {
+              buttons[i].addEventListener('click', function (e) {
+                e.preventDefault();
+                // sendMetric pending implementation of bug 1124471
+                // use bit.ly link to track these clicks
+                // sendMetric('hello-ineligible-view');
+                window.location = '{{ icon_url|safe }}';
+              }, false);
+            }
+          });
+
+          bindHelloObserver();
+        }, false);
+
+        // create namespace
+        if (typeof Mozilla == 'undefined') {
+          var Mozilla = {};
+        }
+
+        'use strict';
+
+        // create namespace
+        if (typeof Mozilla.UITour == 'undefined') {
+          Mozilla.UITour = {};
+        }
+
+        var notificationListener = null;
+
+        function _notificationListener (event) {
+          if (typeof event.detail != 'object')
+            return;
+          if (typeof notificationListener != 'function')
+            return;
+
+          notificationListener(event.detail.event, event.detail.params);
+        }
+
+        function _sendEvent (action, data) {
+          var event = new CustomEvent('mozUITour', {
+            bubbles: true,
+            detail: {
+              action: action,
+              data: data || {}
+            }
+          });
+
+          document.dispatchEvent(event);
+        }
+
+        function _generateCallbackID () {
+          return Math.random().toString(36).replace(/[^a-z]+/g, '');
+        }
+
+        function _waitForCallback (callback) {
+          var id = _generateCallbackID();
+
+          function listener (event) {
+            if (typeof event.detail != 'object')
+              return;
+            if (event.detail.callbackID != id)
+              return;
+
+            document.removeEventListener('mozUITourResponse', listener);
+            callback(event.detail.data);
+          }
+          document.addEventListener('mozUITourResponse', listener);
+
+          return id;
+        }
+
+        Mozilla.UITour.DEFAULT_THEME_CYCLE_DELAY = 10 * 1000;
+        Mozilla.UITour.CONFIGNAME_SYNC = 'sync';
+        Mozilla.UITour.CONFIGNAME_AVAILABLETARGETS = 'availableTargets';
+
+        Mozilla.UITour.ping = function(callback) {
+          var data = {};
+          if (callback) {
+            data.callbackID = _waitForCallback(callback);
+          }
+          _sendEvent('ping', data);
+        };
+
+        Mozilla.UITour.observe = function(listener, callback) {
+          notificationListener = listener;
+
+          if (listener) {
+            document.addEventListener('mozUITourNotification',
+            _notificationListener);
+            Mozilla.UITour.ping(callback);
+          } else {
+            document.removeEventListener('mozUITourNotification',
+            _notificationListener);
+          }
+        };
+
+        Mozilla.UITour.registerPageID = function(pageID) {
+          _sendEvent('registerPageID', {
+            pageID: pageID
+          });
+        };
+
+        Mozilla.UITour.showMenu = function(name, callback) {
+          var showCallbackID;
+          if (callback)
+            showCallbackID = _waitForCallback(callback);
+
+          _sendEvent('showMenu', {
+            name: name,
+            showCallbackID: showCallbackID,
+          });
+        };
+
+        Mozilla.UITour.hideMenu = function(name) {
+          _sendEvent('hideMenu', {
+            name: name
+          });
+        };
+
+        Mozilla.UITour.getConfiguration = function(configName, callback) {
+          _sendEvent('getConfiguration', {
+            callbackID: _waitForCallback(callback),
+            configuration: configName,
+          });
+        };
+
+        function isTargetAvailable (target, loopAvailable, loopNotAvailable) {
+          Mozilla.UITour.getConfiguration('availableTargets', function(config) {
+            if (config.targets && config.targets.indexOf(target) > -1) {
+              loopAvailable();
+            } else {
+              loopNotAvailable();
+            }
+          });
+        }
+
+        function bindHelloObserver () {
+          Mozilla.UITour.observe(function(e) {
+            switch (e) {
+              case 'Loop:ChatWindowOpened':
+                sendMetric('hello-start-conversation');
+                break;
+              case 'Loop:RoomURLCopied':
+                sendMetric('hello-url-copied');
+                break;
+              // bug 1118346
+              case 'Loop:RoomURLEmailed':
+                sendMetric('hello-url-emailed');
+                break;
+            }
+          });
+        }
+
+       function handleVisibilityChange () {
+          if (document.hidden) {
+            // hide Hello menu & stop observer when changing tabs or minimizing window
+            Mozilla.UITour.observe(null);
+          } else {
+            // listen for Hello menu/chat window events
+            bindHelloObserver();
+          }
+        }
+
+      })();
+
+      //]]>
+  </script>
+</div>


templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,237 @@
...
+
+      // create namespace
+      if (typeof Mozilla == 'undefined') {
+        var Mozilla = {};
+      }
+
+      ;(function($) {
+        'use strict';
+
+        // create namespace
```
Osmose: We don't need the entire UITour library included. Take, for example, https://github.com/mozilla/snippets-service/blob/79eaa9aef4caca9329af5780aa1374019e78916c/snippets/base/templates/base/includes/snippet_js.html#L314, where we just create the custom event directly instead of including or relying on the UITour library.

You should be able to create the custom events and add the event listeners you need directly. It'll be less code and easier to read.

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,237 @@
...
+      }
+    </style>
+    
+    {% if icon_url %}<a href="#" id="hello-cta">{% endif %}
+    <img class="icon" src="{{ icon }}" />
+    {% if icon_url %}</a>{% endif %}
+    <p>{{ text|safe }}</p>
+
+    <script type="text/javascript">
+      //<![CDATA[
```
Osmose: This script is executed in the global namespace, so other snippets might be affected by this code accidentally. You want _all_ of the code in this script tag to be inside a closure.

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,237 @@
...
+      var TARGET_AVAILABLE;
+      var TARGET_NAME = 'loop';
```
Osmose: Having a huge list of constants at the start makes the code less readable than just putting the strings directly where they're used.

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,237 @@
+<!-- Any instances of {{ snippet_id }} are replaced with the ID number for the
+     snippet using this template. -->
+
+<div class="snippet" id="{{ snippet_id }}-hello-simple">
+    <style>
```
Osmose: Since you're not using any custom CSS you'll want to remove this block.

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,237 @@
...
+            var targets = config.targets;
+            if (targets && targets.indexOf(target) > -1) {
+              if (track) sendMetric(ELIGIBLE_VIEW);
+              ret = true;
+            }else{
+              if (track) sendMetric(INELIGIBLE_VIEW);
+              ret = false;
+            }
+          });
+          return ret;
```
Osmose: This return shouldn't work; `getConfiguration`'s callback appears to be executed asynchronously, so this function should always return false. You'll want to change this function to take yet another callback that is executed within the callback you pass to `getConfiguration`.

Osmose:
You might want to read up on our JS style guide, there's a few things in here that should be fixed to follow it: http://mozweb.readthedocs.org/en/latest/reference/js-style.html

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,237 @@
...
+        };
+
+        var bindHelloObserver = function() {
```
Osmose: No need to assign these to variables, you can `function bindHelloObserver` fine here.

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,237 @@
...
+
+<div class="snippet" id="{{ snippet_id }}-hello-simple">
+    <style>
+      #{{ snippet_id }}-ffos-circling-fox {
+          /* All CSS should be preceeded by the ID selector for this individual
+            snippet. */
+      }
+    </style>
+    
+    {% if icon_url %}<a href="#" id="hello-cta">{% endif %}
```
Osmose: Use a class instead of an ID here. As per the [snippet development requirements](https://abouthome-snippets-service.readthedocs.org/en/latest/developing.html#snippet-requirements), you should avoid IDs that don't contain the snippet ID, and in general avoid them outside of the root div anyway.

Osmose: Also due to the way you've hooked up the events for this, only the icon will open the hello dialog. If you use a class then an admin can add a link in the snippet text that also opens up the hello dialog: `<a href="#" class="launch-hello">Click here to get started!</a>`.

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,237 @@
...
+              case 'Loop:RoomURLEmailed':
+                sendMetric(URL_EMAILED);
+                break;
+            }
+          });
+        };
+
+       var handleVisibilityChange = function() {
+          if (document.hidden) {
+            // hide Hello menu & stop observer when changing tabs or minimizing window
```
Osmose: Why do you need to hide/stop this stuff?

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,237 @@
...
+          document.addEventListener('visibilitychange', handleVisibilityChange);
+          sendMetric(ELIGIBLE_CLICK);
+        };
+
+        document.getElementById('hello-cta').addEventListener('click', function(e) {
+          e.preventDefault();
+          if (TARGET_AVAILABLE == true) thumbsUp()
+          else {
+            sendMetric(INELIGIBLE_CLICK);
+            window.location.replace('{{ icon_url }}');
```
Osmose: `window.location = '{{ icon_url }}'` should suffice here. Also make sure to test if you need `|safe` appended to the end of this for weird URLs or not.

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,237 @@
...
+
+        Mozilla.UITour.getConfiguration = function(configName, callback) {
+          _sendEvent('getConfiguration', {
+            callbackID: _waitForCallback(callback),
+            configuration: configName,
+          });
+        };
+
+        // custom
+        Mozilla.UITour.isTargetAvailable = function(target, track) {
```
Osmose: This is only ever called with `track=true`, should probably just remove that parameter.

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,237 @@
...
+              TARGET_AVAILABLE = Mozilla.UITour.isTargetAvailable(TARGET_NAME, true);
+          }, false);
+      })();
+
+      // create namespace
+      if (typeof Mozilla == 'undefined') {
+        var Mozilla = {};
+      }
+
+      ;(function($) {
```
Osmose: Leftover `$` from copy-pasting?

Osmose: Also no need for the `;` beforehand, that's usually just a defense by standalone libraries against issues with appending and minifying scripts together if one doesn't use semicolons at the end of its lines.

Osmose: Also also you can just re-use the closure above instead of creating a second one.

Osmose:
Nice first pass! Let me know when you update and I'll go over it again.

templates/hello/simple-snippet.html
ckprice: Assuming you were alluding to the fact that I left some UITour declarations here that I wasn't using in the code. It's been cleaned up.

templates/hello/simple-snippet.html
ckprice: Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1113896#c2

ckprice:
@Osmose messaged you in IRC, but adding comment here as well. Updates made.

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,204 @@
...
+    </a>
+    <p>{{ text|safe }}</p>
+
+    <script type="text/javascript">
+      //<![CDATA[
+
+      (function() {
+          var snippet = document.getElementById('{{ snippet_id }}-hello-simple');
+          snippet.addEventListener('show_snippet', function() {
+              Mozilla.UITour.trackHelloStatus ();
```
Osmose: No space between method name and parenthesis.

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,204 @@
...
+          };
+
+          Mozilla.UITour.getConfiguration = function(configName, callback) {
+            _sendEvent('getConfiguration', {
+              callbackID: _waitForCallback(callback),
+              configuration: configName,
+            });
+          };
+
+          Mozilla.UITour.trackHelloStatus = function () {
```
Osmose: You have the `isTargetAvailable` helper below which can make this function a lot simpler, you just need to have a successCallback and optional failureCallback parameter for that method.

Not to mention, you use that callback below anyway, so there's not much point in checking for the target twice when you can just add the `hello-eligible-view` metric to the callback where you set up the event handlers below.

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,204 @@
...
+            Mozilla.UITour.registerPageID('hello-tour_OpenPanel_snippet-{{ snippet_id }}');
+            // bug 1113896
+            document.addEventListener('click', function(e) {
+                Mozilla.UITour.hideMenu ('loop');
+            }, false);
+            document.addEventListener('visibilitychange', handleVisibilityChange);
+            sendMetric('hello-open');
+          };
+
+          var buttons = document.getElementsByClassName('launch-hello');
```
Osmose: A few things:

1. This selects all buttons with a `launch-hello` classname in the entire document, including other instances of this snippet. You should use the `snippet` DOM element you have above to limit your search to only select elements within this individual snippet.

2. This code executes as soon as the snippet is embedded in the DOM. It's preferred to not execute anything unless this snippet is the one being shown, so this block of code should be moved to the inside of the `show_snippet` event handler.

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,204 @@
...
+          for (var i=0; i < buttons.length; i++) {
+            buttons[i].addEventListener('click', function(e) {
+              e.preventDefault();
+              isTargetAvailable('loop', function(ret){
+                if (ret === true) thumbsUp()
```
Osmose: This `if` statement could be formatted better according the the style guide to be more readable:

```js
if (ret === true) {
  thumbsUp()
} else {
  sendMetric('hello-ineligible-click');
  // ...
}
```

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,204 @@
...
+            }
+          }
+
+          function thumbsUp () {
+            bindHelloObserver();
+            Mozilla.UITour.showMenu ('loop');
+            // bug 1115227
+            Mozilla.UITour.registerPageID('hello-tour_OpenPanel_snippet-{{ snippet_id }}');
+            // bug 1113896
+            document.addEventListener('click', function(e) {
```
Osmose: You'll want to remove this listener once it executes so that you don't stack multiple event listeners and repeatedly call `hideMenu`.

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,204 @@
...
+            if (document.hidden) {
+              // hide Hello menu & stop observer when changing tabs or minimizing window
+              Mozilla.UITour.observe(null);
+            } else {
+              // listen for Hello menu/chat window events
+              bindHelloObserver();
+            }
+          }
+
+          function thumbsUp () {
```
Osmose: `thumbsUp` is a weird name, doesn't really give a good idea of what this function does. I'd almost just take this and the contents of `bindHelloObserver` and put them directly in the code that calls these functions as they're not really so complex as to require separate functions, nor are they reused.

(Actually `bindHelloObserver` is reused but I have another comment about that coming.)

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,204 @@
...
+                case 'Loop:RoomURLEmailed':
+                  sendMetric('hello-url-emailed');
+                  break;
+              }
+            });
+          }
+
+         function handleVisibilityChange () {
+            if (document.hidden) {
+              // hide Hello menu & stop observer when changing tabs or minimizing window
```
Osmose: This doesn't actually hide the hello menu, just stops the observer. And honestly, I don't think you necessarily need to stop the observer unless these events are triggered when the tab is open but the user is on another tab. Do you know if that's the case?

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,204 @@
...
+          };
+
+          var buttons = document.getElementsByClassName('launch-hello');
+          for (var i=0; i < buttons.length; i++) {
+            buttons[i].addEventListener('click', function(e) {
+              e.preventDefault();
+              isTargetAvailable('loop', function(ret){
+                if (ret === true) thumbsUp()
+                else {
+                  sendMetric('hello-ineligible-click');
```
Osmose: So this highlights an issue with `sendMetric`; the HTTP request this method kicks off may not finish before `window.location` changes and the request is aborted. `sendMetric` _should_ be altered to have a callback that can be executed once the request is finished.

In the meantime, why track ineligible clicks? Won't Fabio be able to measure this by adding a parameter to the icon URL, as the only time it's used is in this case?

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,204 @@
...
+            document.addEventListener('click', function(e) {
+                Mozilla.UITour.hideMenu ('loop');
+            }, false);
+            document.addEventListener('visibilitychange', handleVisibilityChange);
+            sendMetric('hello-open');
+          };
+
+          var buttons = document.getElementsByClassName('launch-hello');
+          for (var i=0; i < buttons.length; i++) {
+            buttons[i].addEventListener('click', function(e) {
```
Osmose: Since this is intended for 35+, there's actually a much easier way of handling this, using a click handler and Element.matches:

```js
snippet.addEventListener('click', function(e) {
    if (e.target.matches('.launch-hello')) {
        // event code goes here
    }
});
```

Avoids the for loop and iteration and all that jazz. Yay future!

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,204 @@
...
+            bindHelloObserver();
+            Mozilla.UITour.showMenu ('loop');
+            // bug 1115227
+            Mozilla.UITour.registerPageID('hello-tour_OpenPanel_snippet-{{ snippet_id }}');
+            // bug 1113896
+            document.addEventListener('click', function(e) {
+                Mozilla.UITour.hideMenu ('loop');
+            }, false);
+            document.addEventListener('visibilitychange', handleVisibilityChange);
+            sendMetric('hello-open');
```
Osmose: What is this metric measuring? Looks like the only difference between it and `hello-eligible-view` is that if a user leaves the tab and then comes back it'll get triggered again, right?

templates/hello/simple-snippet.html
Osmose: Also it's a little weird that this function is the only one that's not from the UITour library that you're adding to the UITour object.

ckprice: > You have the `isTargetAvailable` helper below which can make this function a lot simpler

`trackHelloStatus` calls `sendMetric`, which is added globally below the individual snippet JS. Calling at the bottom of the snippet gives me a `ReferenceError`.

> when you can just add the `hello-eligible-view` metric to the callback where you set up the event handlers below.

We need to get this metric on page load, not when someone interacts with the button.

Osmose: If you move the code at the bottom of the snippet to the `show_snippet` event as suggested, then it will run after the function is defined and be able to call `sendMetric`, and also be triggered as soon as the snippet is shown.

ckprice: Okay. What are your thoughts on adding the event listener before we need it? Anything that happens before the button is clicked will not require an event listener. If we're okay adding it before, we can for sure move the `sendMetric` into the `isTargetAvailable` call.

ckprice:
@Osmose thanks for your help today. Updates are in.

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,209 @@
...
+                e.preventDefault();
+                Mozilla.UITour.showMenu('loop');
+
+                // bug 1115227
+                Mozilla.UITour.registerPageID('hello-tour_OpenPanel_snippet-{{ snippet_id }}');
+
+                // bug 1113896
+                document.addEventListener('click', function(e) {
+                    Mozilla.UITour.hideMenu('loop');
+                    document.removeEventListener('click', this);
```
Osmose: Unless I'm misreading [the docs](http://mdn.beonex.com/en/DOM/element.addEventListener.html), shouldn't `this` refer DOM element the listener is attached to? `removeEventListener` requires you to pass the callback function and the same `useCapture` value as you used in `addEventListener`.

templates/hello/simple-snippet.html
```diff
@@ -0,0 +1,209 @@
...
+                }, false);
+
+                // bug exists where teardown tabs will stop .observe
+                // this is to ensure it is re-started
+                document.addEventListener('visibilitychange', handleVisibilityChange);
+
+                sendMetric('hello-open');
+              }, false);
+            }
+          }, function () {
```
Osmose: Tiny readability tip: If you give this function a name, like "loopNotAvailable", it'll be easier for someone reading the code to notice that this is the failure function handler for the `isTargetAvailable` call. I've never actually done that before myself but I think I'll start!

Osmose:
r+wc, this has really come a long way and looks great! Nice work!

templates/hello/simple-snippet.html
ckprice: Actually adding the `removeEventListener` correctly breaks this implementation. Adding the event listener in a callback to the `showMenu` call above is correct.

ckprice:
@Osmose updates made!

PR #30, opened 2015-01-17
Travis testing

indexgap:


Osmose:
Thanks for the PR, but this repo has no tests to run so we don't need travis.

(When Lonnen asked me for a project that could use some Travis I said "Snippets" meaning https://github.com/mozilla/snippets-service, but soon after realized I was mistaken and it already had Travis set up. https://github.com/mozilla/affiliates is set up with Jenkins currently and would be a good candidate to switch to Travis though!)

lonnen:
haha. wow. that's my fault

PR #31, opened 2015-01-26
[fix bug 1115229] Hide menu on window resize (ref bug 1091785)

ckprice:
@Osmose added an additional event listener per Alex's comment in IRC.

### DIFF
--- a/templates/hello/simple-snippet.html
+++ b/templates/hello/simple-snippet.html
@@ -22,10 +22,10 @@
                 e.preventDefault();
                 Mozilla.UITour.showMenu('loop', function () {
                   // bug 1113896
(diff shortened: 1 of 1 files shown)


Osmose:
@alexgibson r?

alexgibson:
This will fire a lot of events - maybe better only to add a one-time `resize` listener when the hello panel is open?

alexgibson:
LGTM, r+

alexgibson:
Oh, hang on - probably want to make sure your click event gets tidied up as well?

If so, make sure to `removeEventListener` for both `click` and `resize` when the panel gets hidden

ckprice:
@alexgibson r?

alexgibson:
Tidy! r+

PR #32, opened 2015-02-10
SMS me a Firefox for Android Snippet.

glogiotatidis:


sms-android/snippet.html
```diff
@@ -0,0 +1,240 @@
...
+        <span>Send me the link</span>
+      </button>
+    </p>
+
+    <hr>
+    <p class="footer">
+      Also available from <a href="http://mzl.la/OgZo6k" rel="external">Google Play</a>.
+    </p>
+    <p class="footer">
+      SMS &amp; data rates may apply. The intended recipient of the SMS must have consented. Learn More.
```
glogiotatidis: We still wait for a link for 'Learn More'

glogiotatidis:
yo @Osmose do you mind taking a look? 

https://snippets.allizom.org/admin/base/snippet/79/

sms-android/snippet.html
```diff
@@ -0,0 +1,240 @@
...
+       display: inline-block;
+       *display: inline;
+       *zoom: 1;
+       border-radius: 0.25em;
+       box-shadow: 0px 1px 0px 0px rgba(0, 0, 0, 0.2), inset 0px -1px 0px 0px rgba(0, 0, 0, 0.3);
+       background-color: #43a6e2;
+       background-color: #277ac1;
+       background-repeat: repeat-x;
+       background-image: -webkit-linear-gradient(top, #43a6e2, #277ac1);
+       /* Safari 5.1+, Chrome 10+ */
```
Osmose: This is Firefox only, you can remove this.

sms-android/snippet.html
```diff
@@ -0,0 +1,240 @@
...
+   }
+
+   #firefox-android-sms-{{ snippet_id }} .button, .button:link, .button:visited {
+       display: inline-block;
+       *display: inline;
+       *zoom: 1;
+       border-radius: 0.25em;
+       box-shadow: 0px 1px 0px 0px rgba(0, 0, 0, 0.2), inset 0px -1px 0px 0px rgba(0, 0, 0, 0.3);
+       background-color: #43a6e2;
+       background-color: #277ac1;
```
Osmose: Two backgorund colors?

sms-android/snippet.html
```diff
@@ -0,0 +1,240 @@
...
+   }
+
+   #firefox-android-sms-{{ snippet_id }} #number {
+       height: 25px;
+       width: 180px;
+       background: #fff;
+       border-color: #b2b2b2;
+       border-style: solid;
+       border-width: 1px;
+       padding: 3.3333333333333335px 6.666666666666667px;
```
Osmose: Why not just 3 and 6?

sms-android/snippet.html
```diff
@@ -0,0 +1,240 @@
...
+
+  <form id="snippet-{{ snippet_id }}-sms-send" method="post" action="https://www.mozilla.org/en-US/firefox/sms/">
+    <img id="right-img" src="{{ image }}">
+    <h1>{{ header }}</h1>
+    <p id="text">{{ text|safe }}</p>
+
+    <p class="input">
+      <input type="tel"
+             maxlength="14"
+             id="number"
```
Osmose: Use a class instead of an ID for this in case multiple instances of this go out.

sms-android/snippet.html
```diff
@@ -0,0 +1,240 @@
...
+             name="number"
+             placeholder="Enter your phone number"
+             value="">
+      <button id="submit-button" type="submit" class="button arrow insensitive" disabled>
+        <span>Send me the link</span>
```
Osmose: Do we want this and the other strings below to be customizable as well? (I'm assuming this is en-US only due to SMS)

sms-android/snippet.html
```diff
@@ -0,0 +1,240 @@
...
+    <p id="text">{{ text|safe }}</p>
+
+    <p class="input">
+      <input type="tel"
+             maxlength="14"
+             id="number"
+             name="number"
+             placeholder="Enter your phone number"
+             value="">
+      <button id="submit-button" type="submit" class="button arrow insensitive" disabled>
```
Osmose: Same with the ID here

sms-android/snippet.html
```diff
@@ -0,0 +1,240 @@
...
+   }
+  </style>
+  <script type="text/javascript">
+   //<![CDATA[
+   (function() {
+       var snippet = document.getElementById('firefox-android-sms-{{ snippet_id }}');
+       snippet.addEventListener('show_snippet', function() {
+           var input = document.querySelector('#firefox-android-sms-{{ snippet_id }} #number');
+           input.addEventListener('input', function() {
+               var button = document.querySelector('#firefox-android-sms-{{ snippet_id }} #submit-button');
```
Osmose: You can cache these once at the start.

sms-android/snippet.html
```diff
@@ -0,0 +1,240 @@
...
+                   button.classList.remove('insensitive');
+               }
+               else {
+                   button.disabled = true;
+                   button.classList.add('insensitive');
+               }
+           });
+           var submit = document.getElementById('submit-button');
+           var form = document.getElementById('snippet-{{ snippet_id }}-sms-send');
+           submit.addEventListener('click', function(event) {
```
Osmose: This only sends the metric if users click the button, not if they directly submit the form. Is it possible to do this on submit of the form?

sms-android/snippet.html
Osmose: This applies to the other stuff like -webkit rules.

Osmose:
The "Send me the link" button should change the cursor to the hand pointer to indicate it's clickable.

Osmose:
Pretty cool, I like how the snippet looks! Lemme know when you make your updates.

sms-android/snippet.html
glogiotatidis: I was thinking that we can probably keep this in case we decide to share the snippets.m.c link on social networks as we've done in the past.

sms-android/snippet.html
glogiotatidis: These are just copy/paste from bedrock css.

sms-android/snippet.html
glogiotatidis: Yeap, en-US due to SMS, no customization.

sms-android/snippet.html
glogiotatidis: Interesting. This is what `lessc` produces for bedrock less (Try compiling `sandstone-resp.less`). cc @craigcook or @alexgibson

alexgibson: Looks like this is because the default button `background-color` property in Snadstone is followed by a vertical gradient mixin, which also adds another `background-color` occurrence as a fallback. This certainly isn't needed, and you can remove the extra property here.

https://github.com/mozilla/bedrock/blob/master/media/css/sandstone/buttons.less#L14

sms-android/snippet.html
alexgibson: These are the compiled values by Less I'm guessing. You can probably just round these down to 1 decimal place, which the browser will do anyway.

sms-android/snippet.html
Osmose: We can't do that with the preview view because it goes away when a snippet is disabled. Also because it uses chrome:// urls for the images and some of the CSS. Also because the about:home code on the preview page is kind've out of date.

If they want to share the snippet on social channels they're going to have to stand up a special static version of it or we're going to have to resolve [bug 1084077](https://bugzilla.mozilla.org/show_bug.cgi?id=1084077).

glogiotatidis:
I'll leave the button's cursor as is to match the behavior of the rest of the buttons on about:home and on the https://www.mozilla.org/en-US/firefox/sms which we copy from.

sms-android/snippet.html
glogiotatidis: ack, thnx!

sms-android/snippet.html
glogiotatidis: Thanks for the detailed explanation Alex!

sms-android/snippet.html
glogiotatidis: Directly == 'by hitting enter while text input is focused'?

Currently the event triggers both on button click and on enter.

Osmose: Holy crap you're right. What?

Is the browser making a fake click event for submit buttons when the form is submitted via the Enter key? That's crazy.

glogiotatidis: ¡~magic~!

glogiotatidis:
@Osmose updated. Can you please take a look first thing in the morning? I need to put this in production before I sign off for the day. Thanks!

glogiotatidis:
Due to time constrains I had to put this on prod. I did all the requested blocking changes and verified with versions 35, 31, 28, 25. Still i'd like your final. thnx!

sms-android/snippet.html
```diff
@@ -0,0 +1,220 @@
...
+         border-radius: 3px;
+         box-shadow: 0 1px rgba(255, 255, 255, 0.5);
+     }
+  </style>
+  <script type="text/javascript">
+   //<![CDATA[
+   (function() {
+       var snippet = document.getElementById('firefox-android-sms-{{ snippet_id }}');
+       snippet.addEventListener('show_snippet', function() {
+           var input = document.querySelector('#firefox-android-sms-{{ snippet_id }} .number');
```
Osmose: Could probably just user `snippet.querySelector` here. Saves on some typing. :D

Osmose:
r+, nice work!

glogiotatidis:
Thanks!

PR #33, opened 2015-02-12
Bug 1120739 Add Hello snippet with animation

sgarrity:
[Bug 1120739](https://bugzilla.mozilla.org/show_bug.cgi?id=1120739)
Test here: https://snippets.allizom.org/show/69/

### DIFF
--- a/templates/hello/animation-snippet.html
+++ b/templates/hello/animation-snippet.html
@@ -1,0 +1,274 @@
+<!-- Any instances of {{ snippet_id }} are replaced with the ID number for the
+     snippet using this template. -->
+<div class="snippet" id="snippet-{{ snippet_id }}">
+  <style>
+    /* All CSS should be preceeded by the ID selector for this individual
+       snippet. */
+    #snippet-{{ snippet_id }}-contents {
+      opacity: 0;
+      position: absolute;
+      bottom: -10px;
+      left: -50px;
+      width: 547px;
+      text-align: center;
+      color: #3c3c3c;
+      transition: opacity .4s ease-in .2s;
+    }
+    #snippet-{{ snippet_id }}-contents p {
+      margin: 1rem 0;
+    }
+    #snippet-{{ snippet_id }}-contents a {
+      background: #00b7e2;
+      color: #fff;
+      display: inline-block;
+      padding: .8rem 2rem;
+      border-radius: .2rem;
+      transition: background ease-in .2s;
+    }
+    #snippet-{{ snippet_id }}-contents a:hover {
+        background: #00d1ef;
+    }
+    #snippet-{{ snippet_id }}-contents a:active {
+        background: #009dd5;
+    }
+    .show-contents-{{ snippet_id }} #snippet-{{ snippet_id }}-contents {
+      opacity: 1;
+    }
+    div.switch-{{ snippet_id }}#brandLogo {
+      position: relative;
+      background-size: auto;
+      height: 192px;
+      width: 447px;
+      transition: background-position .3s ease-in-out;
+    }
+    div.show-contents-{{ snippet_id }}#brandLogo {
+      /* move finished animation up to make room for text and button */
+      background-position: 0 -50px;
+    }
+  </style>
+
+  <div id="snippet-{{ snippet_id }}-contents">
+    <p>{{ text|safe }}</p>
+    <a href="{{ fallbackUrl|safe }}" id="snippet-{{ snippet_id }}-button">
+      {{ buttonLabel|safe }}
+    </a>
+  </div>
+</div>
+
+<script>
+//<![CDATA[
+(function() {
+
+  var snippet = document.getElementById('snippet-{{ snippet_id }}');
+
+  snippet.addEventListener('show_snippet', function() {
+
+    var snippetContents = document.getElementById('snippet-{{ snippet_id }}-contents');
+    var contentDelay = 10000; // Delay before showing text/button
+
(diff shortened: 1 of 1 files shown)


templates/hello/animation-snippet.html
```diff
@@ -0,0 +1,274 @@
...
+      brandLogo.style.backgroundImage = 'url(' + animationImg.src + ')';
+      brandLogo.className += ' switch-{{ snippet_id }}';
+      brandLogo.appendChild(snippetContents);
+      setTimeout(function() {
+        brandLogo.className += ' show-contents-{{ snippet_id }}';
+      }, contentDelay);
+    });
+
+
+    // Begin UITour shinanigins
```
sgarrity: Most of the code below was taken directly, or slightly modified, from this page: https://github.com/mozilla/snippets/blob/master/templates/hello/simple-snippet.html

ckprice:
r+ on all of the UITour Javascript interaction (line 85+).

@glogiotatidis are you comfortable reviewing the CSS and loading of the animation?

glogiotatidis:
Sure I'll review tomorrow.

ckprice:
Thanks Giorgos! Let me know if you run into any issues. Alex said he can help too.

ckprice:
@glogiotatidis please hold on code review. There could be some late changes coming in from PMM.

glogiotatidis:
@ckprice ok, ping again me when ready

templates/hello/animation-snippet.html
```diff
@@ -0,0 +1,274 @@
...
+  var snippet = document.getElementById('snippet-{{ snippet_id }}');
+
+  snippet.addEventListener('show_snippet', function() {
+
+    var snippetContents = document.getElementById('snippet-{{ snippet_id }}-contents');
+    var contentDelay = 10000; // Delay before showing text/button
+
+    // Preload the new animation image from the CDN 
+    var animationImg = new Image();
+    animationImg.src = 'https://snippets.allizom.org/media/files/08f669b5-5e85-4b68-8025-d27f76913ad0.gif';
```
glogiotatidis: Prod url https://snippets.cdn.mozilla.net/media/files/cf92d5d5-3ae5-434d-9b23-ba9ab63a2fd0.gif

glogiotatidis:
ckprice: "Also let's add 5px of top/bottom padding around the copy block"

glogiotatidis:
This one has the wrong bug number.

sgarrity:
> ckprice: "Also let's add 5px of top/bottom padding around the copy block"

@ckprice Can you be more specific about this? We're fairly limited in height.

ckprice:
@sgarrity this was a P2 request from Fabio. If we are limited in vertical space, please ignore.

sgarrity:
I've tried switching back from the CDN to a directly embedded image, but I get a half-second(ish) delay when loading.

If I understand the snippets correctly, I can't hide the Firefox logo by default (until the snippet JS runs and adds a custom CSS class) because the snippet code needs to be completely self-contained.

ckprice:
Summoning @Osmose...

1. Once the CDN was made available to us, my understanding is we should use it to externally load graphics as the default. As opposed to embedding the image into the JS. Mike can you confirm which should be the default?
2. Mike, any recommendation on how to hide the static logo? Showing the Firefox logo for a brief glimpse, then showing the Hello animation looked a bit glitchy. But if this is best practice, we should go with it.

Thanks

Osmose:
> Once the CDN was made available to us, my understanding is we should use it to externally load graphics as the default. As opposed to embedding the image into the JS. Mike can you confirm which should be the default?

I don't think there is one default, they have different uses. Relying on the CDN is useful for large images that bloat the snippet filesize and that would get some benefit out of being cached across locales, whereas embedding the image into the snippet is useful when you want an image to be available immediately when the snippet code executes instead of having to deal with the separate load of an external resource.

> Mike, any recommendation on how to hide the static logo? Showing the Firefox logo for a brief glimpse, then showing the Hello animation looked a bit glitchy. But if this is best practice, we should go with it.

Assume that it is unavoidable to have the Firefox logo visible before the snippet loads. You can make it nice by fading it out instead of just directly replacing it, but you can't get rid of that gap (as the page is already displayed by the time snippet code starts executing).

ckprice:
Thanks Mike!

@glogiotatidis Fabio has asked we bring what we have to production. Can this be ready for him in its current state Monday?

glogiotatidis:
The snippet that we serve is different from the one in this PR. @sgarrity can you please update?

glogiotatidis:
I reviewed the code in the template https://snippets.mozilla.com/admin/base/snippettemplate/31/ and I found one issue:

On page re-load we don't rewind the animation. Instead the animation continues (or if it's done displays the very last frame). I can reproduce this on about:home environment 

Other than that the snippet works fine, nice work @sgarrity

sgarrity:
> The snippet that we serve is different from the one in this PR. @sgarrity can you please update?

Where did you get the version you're using? My intention is to use the version in the PR.

glogiotatidis:
Merged both cdn and embeded versions https://github.com/mozilla/snippets/commit/569da43a31c7a91cc2c7fbbcbc03484b303c22e5  https://github.com/mozilla/snippets/commit/f9be45856cdf18bfed34e8bdc49ecbb049568245 https://github.com/mozilla/snippets/commit/8334f4995ad95880289f1e02d31c91c61620e51c

glogiotatidis:
Thanks @sgarrity !

PR #38, opened 2015-05-13
[fix bug 1140456] Create simple video snippet

ckprice:
@glogiotatidis r?

https://bugzilla.mozilla.org/show_bug.cgi?id=1140456

Inspired from
https://github.com/mozilla/snippets/blob/master/templates/10th-anniversary-video/10th-anniversary-video-no-share.html

### DIFF
--- a/video-snippets/simple-over-search-bar.html
+++ b/video-snippets/simple-over-search-bar.html
@@ -1,0 +1,216 @@
+<style>
+@font-face {
+    font-family: 'fontello-{{ snippet_id }}';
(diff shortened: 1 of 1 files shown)


video-snippets/simple-over-search-bar.html
```diff
@@ -0,0 +1,217 @@
...
+    font-weight: normal;
+    font-style: normal;
+}
+
+#simple-video-snippet-{{ snippet_id }} {
+    width: 470px;
+    margin: 0 auto;
+    display: inline-block;
+    text-align: start;
```
glogiotatidis: i didn't know that, nice!

video-snippets/simple-over-search-bar.html
```diff
@@ -0,0 +1,217 @@
...
+
+        replay.classList.remove('show');
+    });
+
+    // Track progress through the video.
+    var lastProgress = 0;
+    var thresholds = [0.25, 0.5, 0.75, 1];
+    video.addEventListener('timeupdate', function() {
+        var currentProgress = video.currentTime / video.duration;
+        if (currentProgress > lastProgress && !video.paused) {
```
glogiotatidis: So if video has ended, `currentProgress==1` and `video.paused==True` therefore you don't ever send `simple-video-threshold-1`, correct? Is this why you added line 206?

glogiotatidis:
Looks good. Just the metric sending logic is a bit complicated. If we can reftactor this into something simpler it would be great, otherwise we can just go ahead with what we have if it works the way you want.

ckprice:
Thanks for the r @glogiotatidis. Addressed your sendMetric comment. Thoughts?

glogiotatidis:
lgtm, nice work! r+ :cake:

PR #41, opened 2015-06-01
Modify metrics to match naming convention

ckprice:
Fabio/Jean noted that metrics should have a certain naming convention. It will only affect recent templates.

Share snippets
	impression
	conversion-share-facebook
	conversion-share-twitter
Video snippet
	impression
	click-video-play
	video-threshold-[.25,.5,.75,1]
	conversion-video-twitter
	conversion-video-facebook


glogiotatidis:
Thanks!

PR #45, opened 2015-06-16
[fix bug 1163690] Pocket simple snippet

ckprice:
https://bugzilla.mozilla.org/show_bug.cgi?id=1163690

https://snippets.allizom.org/show/86/

@glogiotatidis r?

### DIFF
--- a/templates/pocket/simple.html
+++ b/templates/pocket/simple.html
@@ -1,0 +1,183 @@
+<!-- Any instances of {{ snippet_id }} are replaced with the ID number for the
+     snippet using this template. -->
+
+<div class="snippet" id="{{ snippet_id }}-pocket-simple">
+    <a href="#" class="launch-pocket">
+      <img class="icon" src="{{ icon }}" />
+    </a>
+    <p>{{ text|safe }}</p>
+
+    <script type="text/javascript">
+      //<![CDATA[
+
+      (function() {
+        var snippet = document.getElementById('{{ snippet_id }}-pocket-simple');
+        snippet.addEventListener('show_snippet', function () {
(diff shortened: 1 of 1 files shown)


glogiotatidis:
My only concern is that if user has removed the pocket icon from both the url bar and the drawer, clicking the link does nothing. Maybe we should alter the message based on `isTargetAvailable()`  value?

ckprice:
@glogiotatidis hmm, is it possible for this to happen? I don't think a user can remove the icon completely. Since it's a snippet, we'll only be deploying to locales with the Pocket icon in the build.

ckprice:
Actually, thinking more about this, I'm not sure there is a case where it would do nothing. Except for testing purposes where the `testingOrigins` about:config key hasn't been set.

glogiotatidis:
Actually i've removed the pocket icon from both the url bar and the drawer. You can do this by clicking customize and dragging the icon outside the drawer into the "additional tools and features list" on the left. In this case the snippet does nothing.

I can see people doing this but i don't have numbers on how many. Since we have a way to test this I suggest that we do something, for example display a proper message. If you think otherwise I'll review as is.

ckprice:
@glogiotatidis if I move this to the customization palette as you described, the click action changes to whatever the 'icon_url' the user has configured. In the case of the above sample, it is google.com.

This is not what you experience?

glogiotatidis:
* Changed `icon_url` to `fallback_url`
* Yes I can see the redirect. The reason it didn't work for me before was that I was testing it in the preview mode. Works fine now.

Thanks @ckprice, you rock! :metal:

ckprice:
No, @glogiotatidis, YOU rock!

PR #44, opened 2015-06-10
[bug 1137290] Android SMS and Email snippet

ckprice:
https://bugzilla.mozilla.org/show_bug.cgi?id=1137290

https://snippets.allizom.org/show/90/

PR #46, opened 2015-06-30
[fix bug 1177829] Android SMS/Email with configurable IDs

ckprice:
This is a clone of #44 with the exception of allowing the admin to configure the SMS and Email ID's.

Sample is https://snippets.allizom.org/show/91/
Bug is https://bugzilla.mozilla.org/show_bug.cgi?id=1177829

@glogiotatidis r?


glogiotatidis:
Looks good, thanks @ckprice ! r+

PR #47, opened 2015-07-21
Add language as a template field for Basket email

ckprice:
TODO: somehow use data-locale to do this

@Osmose could you give this a once over with Giorgos out? Note: the template has already been updated on stage/prod.

### DIFF
--- a/android/android-email-sms-configure.html
+++ b/android/android-email-sms-configure.html
@@ -258,13 +258,14 @@
            }
 
            function sendEmailToBasket () {
-                var locale = document.getElementById('snippet_set').getAttribute('data-locale');
+                //TODO: get language from data-locale
+                var lang = '{{ language }}';
                 var url = 'https://basket.mozilla.org/news/subscribe/';
                 var newsletters = '{{ email_id }}';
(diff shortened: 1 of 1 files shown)


Osmose:
One question: Is the problem here that "set of language codes that can be sent to basket" is not the same as "set of locale codes used on the snippet element"?

If the answer is yes, then r+ to this as a temporary measure until a better method of mapping can be found. Otherwise I don't see why `lang` is a separate value.

Osmose:
If basket just takes two-letter codes, maybe `locale.substr(0, 2)` would suffice. Or if you're worried about being able to preview it, maybe add `data-locale` to the preview template? https://github.com/mozilla/snippets-service/blob/master/snippets/base/templates/base/preview.html#L31

ckprice:
> If the answer is yes, then r+ to this as a temporary measure until a better method of mapping can be found. Otherwise I don't see why lang is a separate value.

Per IRC I'd like the ability for an admin to be able to confirm the language of the email they are receiving through preview. Otherwise they wouldn't be able to test the snippet fully.

(please merge when able)

Osmose:
Merged in d1b5ffb

PR #53, opened 2015-09-03
[bug 1200762] Logo swap template blockable and clickable.

glogiotatidis:
Based on logo swap template and the simple snippet with close https://github.com/mozilla/snippets/blob/master/templates/simple-snippet.html

Posted on stage: https://snippets.allizom.org/admin/base/snippettemplate/43/

Note that the close functionality cannot be tested in preview mode.

See also [1172579](https://bugzilla.mozilla.org/show_bug.cgi?id=1172579)

### DIFF
--- a/templates/logo-swap-close.html
+++ b/templates/logo-swap-close.html
@@ -22,7 +22,7 @@
        padding-top: 5px;
        padding-bottom: 5px;
        z-index: -10;
-       border-radius: 9pt;
+       border-radius: 12px;
        visibility: hidden;
        border: solid 0 white;
        box-shadow: 0 0 0 white;
@@ -32,8 +32,8 @@
  {% if clickable %}
    .block-snippet-overlay-glow-{{ snippet_id }} {
        visibility: visible;
-       border: 2pt solid  white;
-       box-shadow: 0 0 10pt #7ac1f9;
+       border: 2px solid  white;
+       box-shadow: 0 0 6px 2px #4CB1FF;
    }
 
    #snippet-{{ snippet_id }}:hover {
--- a/templates/logo-swap-close.html
+++ b/templates/logo-swap-close.html
@@ -24,16 +24,16 @@
        z-index: -10;
        border-radius: 9pt;
        visibility: hidden;
-       border: solid 0pt white;
-       box-shadow: 0pt 0pt 0pt white;
+       border: solid 0 white;
+       box-shadow: 0 0 0 white;
    }
  {% endif %}
 
  {% if clickable %}
    .block-snippet-overlay-glow-{{ snippet_id }} {
        visibility: visible;
-       border: solid 2pt white;
-       box-shadow: 0pt 0pt 10pt #7ac1f9;
+       border: 2pt solid  white;
+       box-shadow: 0 0 10pt #7ac1f9;
    }
 
    #snippet-{{ snippet_id }}:hover {
@@ -74,8 +74,8 @@
    .block-snippet-overlay-close-{{ snippet_id }} {
      visibility: visible;
      background-color: #eaeaea;
-     border: solid 0pt white;
-     box-shadow: 0pt 0pt 0pt white;
+     border: solid 0 white;
+     box-shadow: 0 0 0 white;
    }
  {% endif %}
 </style>
@@ -141,7 +141,7 @@
           brandLogo.insertAdjacentHTML(
             'afterbegin', '<div class="block-snippet-button block-snippet-button-{{ snippet_id }}"></div>');
 
-          var xbuttons = document.getElementsByClassName('block-snippet-button-{{ snippet_id }}');
+          var xbuttons = document.querySelectorAll('.block-snippet-button-{{ snippet_id }}');
           var mouseover = function(event) {
               overlay.classList.add('block-snippet-overlay-close-{{ snippet_id }}');
               brandLogo.style.opacity = 0.6;
@@ -150,7 +150,7 @@
               overlay.classList.remove('block-snippet-overlay-close-{{ snippet_id }}');
               brandLogo.style.opacity = 1;
           };
-          for (var k=0; k < xbuttons.length; k++) {
+          for (var k = 0, l = xbuttons.length; k < l; k++) {
               var xbutton = xbuttons[k];
               xbutton.addEventListener('mouseover', mouseover, false)
               xbutton.addEventListener('mouseout', mouseout, false);
--- a/templates/logo-swap-close.html
+++ b/templates/logo-swap-close.html
@@ -1,0 +1,163 @@
+<style>
+  div.switch-{{ snippet_id }}#brandLogo {
+    background-size: {{ width|default(192, true) }}px {{ height ~ 'px' if height else 'auto' }};
+  }
+
+  .switch-{{ snippet_id }}#brandLogo {
+    height: {{ height|default(192, true) }}px;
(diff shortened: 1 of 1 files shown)


glogiotatidis:
@schalkneethling r?

templates/logo-swap-close.html
```diff
@@ -0,0 +1,163 @@
...
+   .block-snippet-overlay-{{ snippet_id }} {
+       position: absolute;
+       height: 100%;
+       width: 100%;
+       left: 0px;
+       top: 0px;
+       padding-top: 5px;
+       padding-bottom: 5px;
+       z-index: -10;
+       border-radius: 9pt;
```
schalkneethling: I would prefer if we use `px` here as apposed to `points`

templates/logo-swap-close.html
```diff
@@ -0,0 +1,163 @@
...
+       visibility: hidden;
+       border: solid 0pt white;
```
schalkneethling: o values does not need a "type" can just be `0`

templates/logo-swap-close.html
```diff
@@ -0,0 +1,163 @@
...
+       box-shadow: 0pt 0pt 0pt white;
```
schalkneethling: same here

templates/logo-swap-close.html
```diff
@@ -0,0 +1,163 @@
...
+   }
+ {% endif %}
+
+ {% if clickable %}
+   .block-snippet-overlay-glow-{{ snippet_id }} {
+       visibility: visible;
+       border: solid 2pt white;
```
schalkneethling: again, preference for using pixels as apposed to points

templates/logo-swap-close.html
```diff
@@ -0,0 +1,163 @@
...
+   }
+
+   #brandLogo:hover > .block-snippet-button-{{ snippet_id }} {
+       visibility: visible;
+   }
+
+   .block-snippet-overlay-close-{{ snippet_id }} {
+     visibility: visible;
+     background-color: #eaeaea;
+     border: solid 0pt white;
```
schalkneethling: Here as well and below, just `0` will do.

templates/logo-swap-close.html
```diff
@@ -0,0 +1,163 @@
...
+                  links[0].click();
+              }
+          });
+        {% endif %}
+
+        {% if blockable %}
+          brandLogo.insertAdjacentHTML(
+            'afterbegin', '<div class="block-snippet-button block-snippet-button-{{ snippet_id }}"></div>');
+
+          var xbuttons = document.getElementsByClassName('block-snippet-button-{{ snippet_id }}');
```
schalkneethling: `querySelector` will probably be faster.

templates/logo-swap-close.html
```diff
@@ -0,0 +1,163 @@
...
+          var mouseover = function(event) {
+              overlay.classList.add('block-snippet-overlay-close-{{ snippet_id }}');
+              brandLogo.style.opacity = 0.6;
+          };
+          var mouseout = function(event) {
+              overlay.classList.remove('block-snippet-overlay-close-{{ snippet_id }}');
+              brandLogo.style.opacity = 1;
+          };
+          for (var k=0; k < xbuttons.length; k++) {
```
schalkneethling: Bit of an optimization is to cache the array length even though it is really small in this case.

```
for (var k = 0,l = xbuttons.length; k < l; k++) {
```

schalkneethling:
So, after finally getting this up and running locally, thanks @glogiotatidis This works as expected. Shows the snippet, replaced the Fx logo with whatever image you set as the replaced image. Hovering over the snippet presents a close icon, and clicking on this removes the snippet.

schalkneethling:
Does most/all snippets share these rules?

```
.block-snippet-overlay-{{snippet-id}} {
    position: absolute;
    height: 100%;
    width: 100%;
    left: 0px;
    top: 0px;
    padding-top: 5px;
    padding-bottom: 5px;
    z-index: -10;
}
```
If that is the case, we can probably get a lot of reuse and thus less CSS over the wire, if we have a standard class that applies to all and not have it included for each snippet. Perhaps the way snippets are loaded and such prevents us though, but thought I would throw it out there.

schalkneethling:
Not sure where the glow class lives, but I believe (looking at the tiles hover state) that the rules needs to be as follows:

```
.block-snippet-overlay-glow-{{snippet-id}} {
   visibility: visible;
   border: 2px solid fff;
   box-shadow: 0 0 6px 2px #4CB1FF;
}
```

templates/logo-swap-close.html
```diff
@@ -0,0 +1,163 @@
...
+   }
+ {% endif %}
+</style>
+<div id="snippet-{{ snippet_id }}" class="snippet" data-url="{{ logo_url|safe }}">
+  {% if blockable %}
+    <div class="block-snippet-button block-snippet-button-{{ snippet_id }}">
+    </div>
+  {% endif %}
+  {% if blockable or clickable %}
+    <div id="block-snippet-overlay-{{ snippet_id }}" class="block-snippet-overlay-{{ snippet_id }}">
```
schalkneethling: Can we change this to a `a` element or, preferably a `button` element (will need some additional styling to remove the border, background etc.) and then add a `title` attribute with the value `Remove this` or similar.

schalkneethling: The above is partly related to this comment in bugzilla https://bugzilla.mozilla.org/show_bug.cgi?id=1172579#c26 but, also for a11y. Thanks!

templates/logo-swap-close.html
```diff
@@ -0,0 +1,163 @@
...
+       box-shadow: 0pt 0pt 0pt white;
+   }
+ {% endif %}
+
+ {% if clickable %}
+   .block-snippet-overlay-glow-{{ snippet_id }} {
+       visibility: visible;
+       border: solid 2pt white;
+       box-shadow: 0pt 0pt 10pt #7ac1f9;
+   }
```
schalkneethling: You can change the above to:
```
.block-snippet-overlay-glow-{{snippet-id}} {
   visibility: visible;
   border: 2px solid fff;
   box-shadow: 0 0 6px 2px #4CB1FF;
}
```
and it will match tiles exactly.

glogiotatidis:
> If that is the case, we can probably get a lot of reuse and thus less CSS over the wire, if we have a standard class that applies to all and not have it included for each snippet. Perhaps the way snippets are loaded and such prevents us though, but thought I would throw it out there.

Great point. The close button is a new feature so I want to create a couple of templates to see how much is re-usable between them. Then we can share classes between snippets. That includes the close buttons as well.

templates/logo-swap-close.html
glogiotatidis: > Can we change this to an a element or, preferably a button element (will need some additional styling to remove the border, background etc.) and then add a title attribute with the value Remove this or similar.

done.

> The above is partly related to this comment in bugzilla https://bugzilla.mozilla.org/show_bug.cgi?id=1172579#c26 but, also for a11y. Thanks!

Great point again. I want to tackle l10n as well so for the first iteration let's ship this without any text. I updated the bug.

glogiotatidis:
This is updated with the review comments. Multiple commits for your convenience @schalkneethling, I'll squash before merging.

schalkneethling:
This is all good. r+ :hatching_chick:

glogiotatidis:
Thanks!

PR #56, opened 2015-10-05
[bug 1206861] PB Simple Snippet (1of2)

ckprice:
This is snippet 1 of 3 for Private Browsing. See [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1206861) for full specs.

This is dependent on https://github.com/mozilla/snippets-service/pull/122 being pushed to production.

Stage: https://snippets.allizom.org/admin/base/snippet/95/

@glogiotatidis r?

### DIFF
--- a/campaigns/private-browsing/simple.html
+++ b/campaigns/private-browsing/simple.html
@@ -1,0 +1,138 @@
+<style>
+  .snippet .icon {
+    top: 10px;
+  }
+
+ {% if clickable or blockable %}
+   #block-snippet-overlay-{{ snippet_id }} {
+       position: absolute;
+       height: 100%;
+       width: 100%;
+       left: 0px;
+       top: 0px;
+       padding-top: 5px;
+       padding-bottom: 5px;
+       z-index: -10;
+       border-radius: 9pt;
+       visibility: hidden;
+   }
+ {% endif %}
+
+ {% if clickable %}
+   #simple-snippet-{{ snippet_id }}:hover > #block-snippet-overlay-{{ snippet_id }} {
+       visibility: visible;
+       border: solid 2pt white;
+       box-shadow: 0pt 0pt 10pt #7ac1f9;
+   }
+
+   #simple-snippet-{{ snippet_id }}:hover {
+       cursor: pointer;
+   }
+ {% endif %}
+
+ {% if blockable %}
+   #xbutton-{{ snippet_id }} {
+       float: right;
+       z-index: 10;
+       visibility: hidden;
+       height: 28px;
+       width: 28px;
+       margin-right: 5px;
+       margin-top: 5px;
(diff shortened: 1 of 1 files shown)


glogiotatidis:
I see expected launch date 11 Nov so I'll push reviewing those to next week due to bedrock work.

campaigns/private-browsing/simple.html
```diff
@@ -0,0 +1,121 @@
...
+<div class="snippet" id="simple-snippet-{{ snippet_id }}">
+  {% if blockable %}
+    <div class="block-snippet-button" id="xbutton-{{ snippet_id }}">
+    </div>
+  {% endif %}
+  {% if blockable or clickable %}
+    <div id="block-snippet-overlay-{{ snippet_id }}">
+    </div>
+  {% endif %}
+  <a href="{{ icon_url|safe }}" class="launch-pb">
```
glogiotatidis: Please add snippet_id `class="launch-pb-{{ snippet_id }}"`

campaigns/private-browsing/simple.html
```diff
@@ -0,0 +1,121 @@
...
+  <p>{{ text|safe }}</p>
+</div>
+
+{% if clickable %}
+  <script type="text/javascript">
+   //<![CDATA[
+   (function() {
+       var snippet = document.getElementById('simple-snippet-{{ snippet_id }}');
+       snippet.addEventListener('show_snippet', function () {
+          var buttons = snippet.getElementsByClassName('launch-pb');
```
glogiotatidis: Same here `snippet_id`

campaigns/private-browsing/simple.html
```diff
@@ -0,0 +1,121 @@
...
+
+          isTargetAvailable('privateWindow', function targetAvailable () {
+            sendMetric('impression-pb-available');
+            for (var i=0; i < buttons.length; i++) {
```
glogiotatidis: Micro optimization for your loops that caches `buttons.length`:

`for (var i=0, k=buttons.length; i < k; i++) {`

campaigns/private-browsing/simple.html
```diff
@@ -0,0 +1,121 @@
...
+              buttons[i].addEventListener('click', function (e) {
+                e.preventDefault();
+                Mozilla.UITour.showHighlight('privateWindow','wobble');
+                sendMetric('conversion-pb-click');
+              }, false);
+            }
+          }, function targetNotAvailable () {
+            sendMetric('impression-pb-not-available');
+            for (var i=0; i < buttons.length; i++) {
```
glogiotatidis: Same optimization here.

campaigns/private-browsing/simple.html
```diff
@@ -0,0 +1,121 @@
...
+    <div id="block-snippet-overlay-{{ snippet_id }}">
+    </div>
+  {% endif %}
+  <a href="{{ icon_url|safe }}" class="launch-pb">
+    <img class="icon" src="{{ icon }}" />
+  </a>
+  <p>{{ text|safe }}</p>
+</div>
+
+{% if clickable %}
```
glogiotatidis: `clickable` should:
 1) Make the whole snippet (image + text) appear as one big button. Similar to a Tile
 2) Clicking anywhere on the snippet which is now a button should trigger the action, in this case run the code

Right now this snippet does nothing if `clickable` is left unchecked in the snippets admin and if check if fullfils (1) but not (2).

There are two ways to move this forward:
 - If Tile like interface and behavior is not needed for this snippet you can remove `clickable` logic
 - Or otherwise fix it so it executes the code when user clicks the snippet and not only the image of the snippet.

glogiotatidis:
* Snippet blocking works +1
* Snippet logic works +1
 * If Private Window button is available in chrome, clicking on the image highlights the button
 * Otherwise redirects the user to the provided url
* `clickable` logic needs some work.

Nice work @ckprice :)

campaigns/private-browsing/simple.html
ckprice: Ideally we will give the admin the ability to add a link in the snippet text. We've done this in the [past](https://github.com/mozilla/snippets/blob/master/campaigns/pocket/simple.html#L16) (for better or worse). Any recommendations how we can achieve this without risking multiple instances of the same class in the snippet load? For what it's worth, a user should only ever be exposed to one of the private browsing snippets.

campaigns/private-browsing/simple.html
ckprice: See note above.

ckprice:
Thanks @glogiotatidis ! Take another looksy?

campaigns/private-browsing/simple.html
glogiotatidis: :+1: for figuring this out! Sorry for not replying earlier

glogiotatidis:
Merged in 6b2e7f3 with a couple of optimizations and a race condition fix. 

Nice work @ckprice :metal:

PR #58, opened 2015-10-12
[bug 1206861] PB Video (3of3)

ckprice:
This is snippet 3 of 3 for Private Browsing. See bug for full specs.

Stage: https://snippets.allizom.org/admin/base/snippet/97/

@glogiotatidis r? Also, is there some kind of throttling with the video assets? Seems like every 10 minutes or so the video won't load for me. Any idea?

### DIFF
--- a/campaigns/private-browsing/video-over-searchbar.html
+++ b/campaigns/private-browsing/video-over-searchbar.html
@@ -1,0 +1,251 @@
+<style>
+@font-face {
+    font-family: 'fontello-{{ snippet_id }}';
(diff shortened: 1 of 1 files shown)


glogiotatidis:
Similar comments about `launch-pb` selectors, micro-optimization of loops

glogiotatidis:
> Seems like every 10 minutes or so the video won't load for me.

Every 10 minutes? Did you mean tries / reloads?

Videos should get uploaded to the CDN (using `Uploaded files` on the admin) and there's no throttling afaik. Do you have a url I can check?

glogiotatidis:
in gdocs I see that this is cancelled. Is that correct @ckprice ?

ckprice:
Yes, sorry @glogiotatidis =/

PR #61, opened 2015-11-11
 EOY 2015 Preserve existing url params 

ScottDowne:
… prop

### DIFF
--- a/campaigns/mofo-eoy-2015/snippet.html
+++ b/campaigns/mofo-eoy-2015/snippet.html
@@ -225,7 +225,7 @@
         {{ donationAmountFourth }} {{ currencySymbol|default("$", true) }}
       </button>
 
-      <a class="button button-red button-wide donate-button  {{ buttonAnimate|default("", true) }}" target="_blank" href="#">
+      <a class="button button-red button-wide donate-button  {{ buttonAnimate|default("", true) }}" href="#">
         {{ donateButtonText }}
       </a>
     </div>
@@ -250,15 +250,21 @@
           snippet.addEventListener('show_snippet', function() {
               (function () {
                 document.querySelector("body").classList.add("background-{{ snippet_id }}");
-                var donationURL = '{{ donationFormURL }}';
+                var donationURL = '{{ donationFormURL|safe }}';
                 var currencyCode = '{{ currencyCode|default('usd', true) }}';
                 var extraParams = 'snippet_name={{ snippet_id }}';
+                if (donationURL.indexOf('?') >= 0) {
+                    donationURL += '&' + extraParams;
+                } else {
+                    donationURL += '?' + extraParams;
+                }
 
                 // Element cache :
 
                 var btnAmounts = snippet.querySelectorAll('.donation-amount');
                 var btnSelected;
                 var amount;
+                var presets = '{{donationAmountFirst}},{{donationAmountSecond}},{{donationAmountThird}},{{donationAmountFourth}}';
                 var btnDonate = snippet.querySelector('.donate-button');
                 var elHighlighted = snippet.querySelector('em');
                 var monthlyCheckbox = document.querySelector('#monthly-checkbox');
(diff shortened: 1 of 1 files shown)


ScottDowne:
This allows for the config prop `donationFormURL` to already have query string params pre filled, and the snippet knows to handle that, and if it doesn't the snippet still knows.

It also makes sure we default to the same preset values to match the snippet even if frequency is changed. Example, if you change frequency to monthly, the amounts preselected match the snippet at 20,10,5,3, but if you change it to a single donation, it was defaulting it to the wrong amounts.

@glogiotatidis quick review?

campaigns/mofo-eoy-2015/snippet.html
```diff
@@ -250,15 +250,21 @@
           snippet.addEventListener('show_snippet', function() {
               (function () {
                 document.querySelector("body").classList.add("background-{{ snippet_id }}");
-                var donationURL = '{{ donationFormURL }}';
+                var donationURL = '{{ donationFormURL|safe }}';
                 var currencyCode = '{{ currencyCode|default('usd', true) }}';
                 var extraParams = 'snippet_name={{ snippet_id }}';
```
glogiotatidis: snippet_name is automatically added by the service. Do not include here.

glogiotatidis:
I noticed that you set `target=_blank` here https://github.com/ScottDowne/snippets/blob/65517b7d8cddf360c8ffcb4ffb8caa4af98419af/campaigns/mofo-eoy-2015/snippet.html#L228 which is blocked by default in fx and the "the page wants to open a popup" top bar appears. Let's change that.

glogiotatidis:
Please set a proper capitalized commit message.

ScottDowne:
Thanks for the review, working on these now.

ScottDowne:
@glogiotatidis Fixed :)

campaigns/mofo-eoy-2015/snippet.html
```diff
@@ -250,15 +250,20 @@
           snippet.addEventListener('show_snippet', function() {
               (function () {
                 document.querySelector("body").classList.add("background-{{ snippet_id }}");
-                var donationURL = '{{ donationFormURL }}';
+                var donationURL = '{{ donationFormURL|safe }}';
                 var currencyCode = '{{ currencyCode|default('usd', true) }}';
-                var extraParams = 'snippet_name={{ snippet_id }}';
```
glogiotatidis: Apologies you'll have to keep this and ignore my previous comment. The service does indeed set `snippet_name` but the way you manipulate the params when the buttons are clicked it gets removed again.

glogiotatidis:
re commit message I think something along the lines `EOY 2015 Preserve existing url params` is better :)

glogiotatidis:
This is good to go, lets bring back snippet_name and I'll merge and put to prod. Thanks @ScottDowne  :octopus:

ScottDowne:
Updated.

One question, I needed to add |safe to the url so the & in the url doesn't get converted to `&amp;`

But I noticed the params are added as "https://donate.mozilla.org/de/?utm_source=desktop-snippet&amp;utm_medium=snippet&amp;utm_term=5465&amp;utm_campaign=EOY" with a `&amp;` coded into it. Does that make sense and is this going to come end up how we want it once it's shipped?

glogiotatidis:
> But I noticed the params are added as "https://donate.mozilla.org/de/?utm_source=desktop-snippet&amp;utm_medium=snippet&amp;utm_term=5465&amp;utm_campaign=EOY" with a &amp; already in it. Does that make sense and is this going to come end up how we want it once it's shipped?

Good question. When you say `params are added` you mean that the snippet editor pastes a link with `&amp;` already in it, correct?

glogiotatidis:
> Does that make sense and is this going to come end up how we want it once it's shipped?

I just verified this. It will come up with `&amp;` but that's ok since that's the way to do it in xhtml and about:home is xhtml

campaigns/mofo-eoy-2015/snippet.html
```diff
@@ -225,7 +225,7 @@
         {{ donationAmountFourth }} {{ currencySymbol|default("$", true) }}
       </button>
 
-      <a class="button button-red button-wide donate-button  {{ buttonAnimate|default("", true) }}" target="_blank" href="#">
+      <a class="button button-red button-wide donate-button  {{ buttonAnimate|default("", true) }}" href="#">
```
glogiotatidis: btw we are now expecting templates to be valid XML so this would raise an error while saving. I changed it to ` {% if buttonAnimate %} {{ buttonAnimate }} {% endif %}`

glogiotatidis:
Thanks!

campaigns/mofo-eoy-2015/snippet.html
ScottDowne: Works for me :)

PR #63, opened 2015-11-16
If template is clickable wrap around an <a>.

glogiotatidis:
Avoiding using custom 'click' listeners will make handling of click
metrics easier.


glogiotatidis:
@schalkneethling r?

schalkneethling:
Just wondering if it is "ok" to have this be `master` to `master` and not from a specific feature branch. Other than that, r+

PR #64, opened 2015-11-16
 EOY 2015 anchor to work with &amp; in the href 

ScottDowne:


glogiotatidis:
@schalkneethling can you shed some light here? As you know `about:home` is xhtml and we can a link in this template that looks like this

`https://donate.mozilla.org/de/?utm_source=desktop-snippet&amp;utm_medium=snippet&amp;utm_term=5465&amp;utm_campaign=EOY15`

I understand that `&amp;` is the correct way to put `&` in a xhtml page and it should reach `donate.mozilla.org` in the correct form. Is that correct? Do we need what Scott is doing? Is it ok to use JS to add invalid xhtml elements in the page?

schalkneethling:
@glogiotatidis Yes, technically having the ampersands escaped is absolutely fine and the URL should reach it's intended destination just fine. I did run into a problem with these exact URLs where using `&amp;` caused the parameters to not be interpreted correctly so, perhaps it is safe in this case to go with the JS workaround. Especially if it was needed previously.

glogiotatidis:
> Especially if it was needed previously.

We don't have a history with this workaround. AFAIK all snippet links are `&amp;`. 

Can any or both of you give more info about the problems with these urls? I did test them and I did read the parameters correctly on the server side.

ScottDowne:
When I test it, locally, the brief moment it was prod, and when I am shared the results from staging, the link I was getting redirected to had the &amp in them, which isn't going to be parsed correctly.

It's interesting that last year's had this workaround, which is why I thought it was needed.

Can I see an example of where this is working?

glogiotatidis:
> It's interesting that last year's had this workaround, which is why I thought it was needed.

aha ok good to know

> Can I see an example of where this is working?

I just tried it with this template code and custom [bottle](http://bottlepy.org/docs/dev/index.html) app and it worked like a charm

ScottDowne:
@glogiotatidis I'm more than happy to not have my change, but can you confirm first that the `&amp;` is in fact just an `&` after the url redirects?

ScottDowne:
@glogiotatidis also, this brings me to another question. I'm adding in params manually, which I use regular `&` and not `&amp;` would that be the same issue? Maybe mixing them is causing the issue?

glogiotatidis:
> @glogiotatidis also, this brings me to another question. I'm adding in params manually, which I use regular & and not &amp; would that be the same issue? Maybe mixing them is causing the issue?

that's an interesting point. do you mind testing this out? Also how do you test?

glogiotatidis:
> @glogiotatidis I'm more than happy to not have my change, but can you confirm first that the &amp; is in fact just an & after the url redirects?

Yes, with what I tried the bottle app parsed correctly the query arguments.

ScottDowne:
@glogiotatidis I'm testing i locally in a node app, and I am aware that it wouldn't be a reliable example of what's happening on staging or production, which is why I didn't initially push very hard about this being broken.

However, the staging links I am shared are retaining the `&amp;` so I started to consider that it might not just be me.

ScottDowne:
I'm looking at the snippet home page, and the about mozilla link at the top has the `&amp;` in the source, but when I click it, I get the `&` as expected. So I think you are on to something, but I'm going to do more poking around to understand it better.

glogiotatidis:
> I'm looking at the snippet home page, and the about mozilla link at the top has the &amp; in the source, but when I click it, I get the & as expected. So I think you are on to something, but I'm going to do more poking around to understand it better.

Please do, I'm interested in your findings.

ScottDowne:
@glogiotatidis 

Interesting, so I made the `<a href="https://donate.mozilla.org/de/?testone=test&amp;testtwo=test"></a>`

and is returns with a working `&` in the url.

Once I add it via javascript, it breaks. I think the `&amp;` parsing happens before javascript. If I view source on the page, I see the `&amp;` but if I inspect it at run time, it's already converted to a `&`

ScottDowne:
Based on this, I think seeing how I am using the url inside javascript alone, it might be optimal to never have the `&amp;` in the url in the first place.

ScottDowne:
I think I'm making a new pull request with my findings, solving this mystery.

ScottDowne:
@glogiotatidis fixed! Updating commit message, but I think this fix is going to make us both happy.

glogiotatidis:
Works just fine. Many thanks for your research and fix Scott.

ScottDowne:
@glogiotatidis and thanks for your patience :)

PR #65, opened 2015-11-17
EOY 2015 allow for space and ability to display different currency ty…

ScottDowne:
…pes in amount buttons

### DIFF
--- a/campaigns/mofo-eoy-2015/snippet.html
+++ b/campaigns/mofo-eoy-2015/snippet.html
@@ -37,7 +37,7 @@
 
       #mofo-eoy-fundraiser-{{ snippet_id }} .button {
         margin: 0 8px 0 0;
-        width: 57px;
+        width: 60px;
         height: 38px;
         text-align: center;
         padding: 7px 0 9px;
@@ -71,7 +71,7 @@
         background-color: #FF7B87;
       }
       #mofo-eoy-fundraiser-{{ snippet_id }} .button-wide {
-        width: 170px;
+        width: 158px;
       }
 
       #mofo-eoy-fundraiser-{{ snippet_id }} .monthly-checkbox-container {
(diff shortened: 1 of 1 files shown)


campaigns/mofo-eoy-2015/snippet.html
```diff
@@ -266,6 +258,15 @@
                 var btnSelected;
                 var amount;
                 var presets = '{{donationAmountFirst}},{{donationAmountSecond}},{{donationAmountThird}},{{donationAmountFourth}}';
+                var numberFormat = new Intl.NumberFormat('{{locale}}', {
```
glogiotatidis: `Intl.NumberFormat`, nice! this works for Fx 29+ though, we need to verify that this is OK with the team. I'll comment in the bug.

glogiotatidis: `locale` needs a default value, especially because we already run the snippet and any changes must be backwards compatible

glogiotatidis: also btw `{{ locale }}` with spaces between is more readable and consistent with the rest variables.

glogiotatidis:
A couple of nits but before merging we need to verify that Fx >= 29.0 is OK and build a client match rule to filter out the rest.

campaigns/mofo-eoy-2015/snippet.html
ScottDowne: I can fix this asap.

ScottDowne:
@glogiotatidis updated, I went with `{{ locale|default("en-US", true) }}`

ScottDowne:
From the bugzilla issue, less than 1% is below ff 29, so we can not worry :)

glogiotatidis:
Looks good, thanks!

PR #72, opened 2015-12-22
 EOY 2015 add text highlight to spectacular snippet 2 and 3, merge he…

ScottDowne:
…ader and body text in spectacular snippet 2 and 3, and add a text alignment variable to all spectacular snippets

### DIFF
--- a/campaigns/mofo-spectacular-1-eoy-2015/snippet.html
+++ b/campaigns/mofo-spectacular-1-eoy-2015/snippet.html
@@ -91,7 +91,7 @@
       }
 
       #mofo-eoy-fundraiser-{{ snippet_id }} .background-container {
-        text-align: center;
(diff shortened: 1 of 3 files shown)


ScottDowne:
@glogiotatidis Some more updates. We had issues fitting in the translated copy for spectacular snippets 2 and 3, and the highlight functionality that exists in the standard EOY 2015 snippet was requested in 2 and 3. It's the same code that we use previously so shouldn't be problematic.

There was also talk about alignment left for longer blocks of text, so I made that configurable.

Thanks again!

glogiotatidis:
You really like loooong commit messages! :)

Merged https://github.com/mozilla/snippets/commit/7f53e08cf90dc417c678df174eaa40f608dd15ab and updated on production. 

Thanks @ScottDowne :gift:

PR #75, opened 2015-12-23
EOY 2015 ensure we can still scroll-y

ScottDowne:


ScottDowne:
So we noticed in two of our tests, I broke the vertical scroll. This fixes it.

ScottDowne:
This issues is I had to add elements to the fox animation, which caused a horizontal scrollbar. To fix that, I hid the overflow(removed the scrollbar), but didn't cross my mind at the time that it would also hide the vertical scrollbar, which was not my intent.

My fix ensures only the horizontal scrollbar is hidden.

ScottDowne:
To see the bug in action, go here: https://snippets.mozilla.com/show/5508/

And shorten your window height a bit until something gets cut off. You don't see a scrollbar as expected.

ScottDowne:
@schalkneethling review? thoughts?

ScottDowne:
@schalkneethling 

I noticed a few other not great situations when the screen's width gets smaller. I created a separate patch for that though because it was larger than this one, and probably not as critical.

You can find it here: https://github.com/mozilla/snippets/pull/76

Osmose:
r+, merged to fix the issue, but I think in the long term any CSS that messes with the body should be _very_ conservative about things like hiding scrollbars.

PR #76, opened 2015-12-28
Small display

ScottDowne:


ScottDowne:
This does a few things when the screen width gets below 695px.

1. Moves the close button into the container.
2. Update close button padding so it better fits into the new space.
3. Hide the fox, as there isn't enough space to display it anyway.
4. Re-enable the scroll x because the fox is hidden now, we can do this.
5. Shorten the container's width so it can fit into 695px

You can test an example of the intended result for them here:
http://scottdowne.github.io/eoy-snippets/2015-spectacular-snippet-2/
http://scottdowne.github.io/eoy-snippets/2015-spectacular-snippet-3/

@schalkneethling thoughts? review?

glogiotatidis:
Closing this one due to lack of activity.

PR #77, opened 2016-01-14
[fix bug 1238092] Prevent search box from taking focus.

glogiotatidis:


glogiotatidis:
@schalkneethling r?

campaigns/mobile/mobile-snippet-widget-200.html
```diff
@@ -327,6 +327,11 @@ <h2 class="thank-you hidden" id="thank-you-message-{{ snippet_id }}">{{ thank_yo
          var button = document.querySelector('#firefox-mobile-activation-{{ snippet_id }} .submit-button');
          var input_type;
 
+         // Bug 1238092. Can be removed when Bug 1238092 gets fixed.
+         input.addEventListener('keypress', function(event) {
+             event.stopPropagation();
```
schalkneethling: ah, so we just want these keypress event to not bubble and thus not hit the core Fx code that would trap it and send it to the search bar. Correct? Then after whichever form of a fix lands in release, we can remove this? Although, it will then likely still be broken in older versions so, we will probably want to have this hanging around for a while.

glogiotatidis: Exactly as you say: captures the keypress and yes we'll have to keep this around for a few months

schalkneethling:
r+ :coffee:

glogiotatidis:
In retrospective I'll push this change to the service so the issue gets fixed for all the snippets. I'll push this template to production until I land a fix on the server and then I'll revert. 

@schalkneethling do you see any reason to avoid having this in https://github.com/mozilla/snippets-service/blob/master/snippets/base/templates/base/includes/snippet_js.html ?

schalkneethling:
Nope, we need it for all snippets that has an input field so, making it global is probably best. Only thing I am wondering about is, what happens if there is no `input` element i.e. `input` is undefined. I am not 100% sue of the top of my head but, I reckon doing input.addEventListener will throw an error.

If this is not the case, then we should all good to make this global. In fact, we can still make it global, we just need to wrap it in something like:

```
if (input) {
  // add event listener
}
```

PR #80, opened 2016-06-03
Add snippet template for prompting users with a Heartbeat survey.

Osmose:
We're hoping to use this to send out a survey that we originally intended to send via self-repair. Problem is that self-repair can't detect funnelcake builds, and we need to target it to funnelcake builds. Snippets gets the funnelcake ID in the release channel field and thus can target them, although the users will have to hit about:home to see it. But it's the best we can do on short notice.

Email incoming with more details, but I tested this locally with the template dev server and it seems to work fine. I'll do more comprehensive testing on each of the funnelcake builds once this gets merged and is available on staging.

@glogiotatidis r?

### DIFF
--- a/generic-templates/heartbeat-survey.html
+++ b/generic-templates/heartbeat-survey.html
@@ -1,0 +1,101 @@
+{#
+Snippet for prompting users with a UITour survey
+
+Variables:
+
+  @message - small text - Survey message text
+  @thankyouMessage - small text - Thanks message text shown after the user
+    interacts with the prompt
+  @engagementURL - URL - URL to open in a new tab after the user interacts with
+    the prompt
+  @engagementButtonLabel - small text - If blank, show a 5-star rating in the
+    prompt, otherwise show a button with this text
+  @learnMoreURL - URL - URL to open when the Learn More link is clicked
+  @learnMoreLabel - small text - Label for the Learn More link
+
+#}
+<div class="snippet" id="surveysnippet-{{ snippet_id }}"></div>
+<script type="text/javascript">
+//<![CDATA[
+(function() {
+  var ID = 'surveysnippet-{{ snippet_id }}';
+  var snippet = document.getElementById(ID);
+
+  snippet.addEventListener('show_snippet', function() {
+    addToBlockList({{ snippet_id }});
+    annotateUrl('{{ engagementURL|escapejs|safe }}').then(function(engagementURL) {
+      document.dispatchEvent(new CustomEvent('mozUITour', {
+        bubbles: true,
+        detail: {
+  	      action: 'showHeartbeat',
+  	      data: {
+            message: '{{ message|escapejs|safe }}',
+            thankyouMessage: '{{ thankyouMessage|escapejs|safe }}',
+            flowId: 'fake-flowid-snippet-survey',
+            engagementURL: engagementURL,
+            learnMoreLabel: '{{ learnMoreLabel|escapejs|safe }}',
+            learnMoreURL: '{{ learnMoreURL|escapejs|safe }}',
+            engagementButtonLabel: '{{ engagementButtonLabel|escapejs|safe }}'
+          }
+        }
+      }));
+    });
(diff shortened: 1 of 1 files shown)


Osmose:
Whoops, Matt_G has asked that we carry over the post-answer URL annotation. Will be updating this soon.

generic-templates/heartbeat-survey.html
```diff
@@ -0,0 +1,49 @@
...
+  snippet.addEventListener('show_snippet', function () {
+    if (gSnippetsMap.get(ID) !== 'shown' || gSnippetsMap.get('surveysnippet-override')) {
+      gSnippetsMap.set(ID, 'shown');
+
+      var uitourEvent = new CustomEvent('mozUITour', {
+        bubbles: true,
+        detail: {
+  	      action: 'showHeartbeat',
+  	      data: {
+            message: '{{ message }}',
```
jpetto: Nit - indentation.

generic-templates/heartbeat-survey.html
```diff
@@ -0,0 +1,49 @@
...
+            thankyouMessage: '{{ thankyouMessage }}',
```
jpetto: Are these strings expected ready to be placed in JS? Are we worried about unescaped single quotes or anything?

Osmose: Actually without `|safe` I think these will get escaped, and I don't think I want that. At least, in the other snippet templates everything has `|safe` applied, so if there's escaping that needs to be done I think it's done elsewhere. I remember we used to have issues like that and solved them somehow but it's been a long time.

Osmose:
Updated, the post-interaction URL will now be annotated with some extra info about the user as we normally do in the SHIELD surveys.

generic-templates/heartbeat-survey.html
```diff
@@ -0,0 +1,104 @@
+{#
+Snippet for prompting users with a UITour survey
+
+Variables:
+
+  @message - small text - Survey question text
```
glogiotatidis: nit: update docs to match variable names.

generic-templates/heartbeat-survey.html
```diff
@@ -0,0 +1,104 @@
...
+#}
+<div class="snippet" id="surveysnippet-{{ snippet_id }}"></div>
+<script type="text/javascript">
+//<![CDATA[
+(function() {
+  var ID = 'surveysnippet-{{ snippet_id }}';
+  var snippet = document.getElementById(ID);
+
+  snippet.addEventListener('show_snippet', function() {
+    if (gSnippetsMap.get(ID) !== 'shown' || gSnippetsMap.get('surveysnippet-override')) {
```
glogiotatidis: You should use the service blocklist, it's simpler and doesn't pollute the gSnippetsMap namespace:

```javascript
 snippet.addEventListener('show_snippet', function() {
      addToBlockList({{ snippet_id }});
      annotateUrl('{{ engagementURL|safe }}').then(function(engagementURL) {
          ....
```

You'll lose the override option which anyway needs the user to type things in the console to enable it, so they can clear the service blocklist as well.

generic-templates/heartbeat-survey.html
```diff
@@ -0,0 +1,104 @@
...
+      gSnippetsMap.set(ID, 'shown');
+
+      annotateUrl('{{ engagementURL|safe }}').then(function(engagementURL) {
+        document.dispatchEvent(new CustomEvent('mozUITour', {
+          bubbles: true,
+          detail: {
+    	      action: 'showHeartbeat',
+    	      data: {
+              message: '{{ message|safe }}',
```
glogiotatidis: Indeed you need to escape this as @jpetto pointed:
`{{ message|escapejs|safe }}`

glogiotatidis:
Tested and works! Besides the couple of nits this is good to go. Great work as always @Osmose :candy:

Osmose:
Thanks!

PR #82, opened 2016-08-20
Adding a button link snippet.

ScottDowne:
Hey @glogiotatidis 

I've put together a snippet to cover the cases in this document: https://redpen.io/xh8ee647b94e52b73c

It adds a link button, which is added to the snippet if you add a link_button_text prop.

It also allows a variable width icon to handle this: https://redpen.io/fc8451c0686ff1ebb9

I changed the layout to use css flexbox, which was the best way to accomplish variable icon width, with or without  a button, and allows for the snippet to be variable length too.

I thought this might warrant adding these updates to the base snippet, as I basically just modified it, but I didn't want to do that initially, but I'm open to doing that if you think it's a good idea.

Thoughts?

### DIFF
--- a/generic-templates/simple-snippet.html
+++ b/generic-templates/simple-snippet.html
@@ -1,5 +1,5 @@
 {#
-Simple and Logo Swap Snippet with Block and Click
+Simple and Logo Swap Snippet with Block, Link Button and Click
 
 Variables:
 
@@ -10,6 +10,7 @@ Variables:
   @icon_url: text - URL linked from icon and whole snippet when snippet is clickable. (optional / required when snippet is clickable)
   @link_logo: checkbox - Check to make replacement logo clickable.
   @replacement_logo: image - Optional. Replace logo.
+  @snippet_button_text: small text - Optional. Text for a button that links to icon_url.
   @text: main text - Text of snippet.
   @logo_height: small_text - Logo height. Defaults to 192px
   @logo_width: small_text -  Logo width. Defaults to 192px
@@ -78,6 +79,35 @@ div.switch#brandLogo {
  }
 {% endif %}
 
+ .flex-container {
+     display: flex;
+     align-items: center;
+     width: 100%;
+ }
+
+ .flex-start {
+     justify-content: flex-start;
+ }
+
+ .flex-space-between {
+     justify-content: space-between;
+ }
+
+ .button-link {
+     white-space: nowrap;
+     margin: 5px;
+     background: #D3D9DD;
+     color: #000;
+     font-size: 16px;
+     padding: 5px 13px;
+     border-radius: 6px;
+ }
+
+ .button-link:hover {
+     background: #2B9EEF;
+     color: #FFF;
+ }
+
  {% if blockable %}
    .block-snippet-button {
        position: absolute;
@@ -127,17 +157,23 @@ div.switch#brandLogo {
     {% if blockable or clickable %}
       <div id="block-snippet-overlay" class="block-snippet-overlay">
     {% endif %}
+    <div class="flex-container flex-start">
+      {% if icon_url %}
+        <a href="{{ icon_url|safe }}">
+      {% endif %}
+            <img class="icon" src="{{ icon }}" />
+      {% if icon_url %}
+        </a>
+      {% endif %}
 
-    {% if icon_url %}
-      <a href="{{ icon_url|safe }}">
-    {% endif %}
-          <img class="icon" src="{{ icon }}" />
-    {% if icon_url %}
-      </a>
-    {% endif %}
+      <div class="flex-container flex-space-between">
+        <p>{{ text|safe }}</p>
 
-    <p>{{ text|safe }}</p>
-
+        {% if snippet_button_text %}
+        <a class="button-link" href="{{ icon_url|safe }}">{{snippet_button_text}}</a>
+        {% endif %}
+      </div>
+    </div>
     {% if blockable or clickable %}
       </div>
     {% endif %}


glogiotatidis:
Hi Scott

> I've put together a snippet to cover the cases in this document: https://redpen.io/xh8ee647b94e52b73c

Can you provide a document with the cases because the only thing I get from redpen are designs. If the link actually includes the cases please point me to them b/c i'm unfamiliar with redpen.

generic-templates/button-snippet.html
```diff
@@ -0,0 +1,264 @@
+{#
+Simple and Logo Swap Snippet with Block and Click
```
glogiotatidis: Update documentation, variables and provide sample values for each one of them.

ScottDowne:
@glogiotatidis Yeah, the designs are all I got too. I'll ask for more. I'll fix up the snippet docs. Did you see anything wrong with the code?

@hpauljohnson We need some sort of document with use cases about the new snippets we're proposing. It's likely this already exists, it just wasn't put into https://github.com/MozillaFoundation/Advocacy/issues/246

ScottDowne:
@glogiotatidis so I updated this line to mention the change: https://github.com/mozmar/snippets/pull/82/files#diff-6dc6f9f6c62671bc5045f7e72f424ef3R2

I rebased.

I previously had a line with the new variable here: https://github.com/mozmar/snippets/pull/82/files#diff-6dc6f9f6c62671bc5045f7e72f424ef3R12

I'm not sure where you want me to put sample values, though, as I'm just following the pattern in other snippet templates, and not seeing any sample values (yet, maybe just not seeing it), just variable descriptions.

jkt230:
@hpauljohnson - If you can send us the user story behind this as well as the overall test plan (i.e. why are we looking to push this type of snippet template out?) that would be helpful. Thanks.

HPaulJohnson:
@jkt230 The basic user story is EU Firefox users who visit the home page and are interested in participating in our campaign to reform copyright in Europe. The CTA for all versions will be to click through to our petition landing page where we will provide context about the campaign, including 3 simple things we can do to reform copyright laws in the EU to protect basic user rights on the Internet and ask them to sign our petition/pledge. 

On the landing page we are testing a few different message frames that take a few different approaches: 1) more hard-hitting, pointing out the absurdities of current laws, 2) more positive, focused on protecting creativity/imagination online and 3) using some current examples of where copyright laws impact everyday internet users. 

Based on some previous testing done on a donation campaign in Germany earlier this year, we saw by embedding buttons as part of our CTA we were able to double click through rates to our landing page. Leveraging these insights we'd like to test something similar on this campaign, again using a button instead of a hyperlink as the primary CTA.

Let me know if that's sufficient context or if you'd like additional background.

jkt230:
Hi @HPaulJohnson and @ScottDowne 

Thanks for the response. I've also checked with the UX guidelines to make sure it's aligned and it looks like it'll be okay to test given we don't 1) move up or down the static elements of the page and 2) include an opt-out option. (Guidelines here: https://docs.google.com/document/d/1-JWdwTtSEyHCde2xcmEzURD7EBh45G_M4Hg5c7KWDVU/edit)

@glogiotatidis I think with following the guidelines and the use case story we're good to go but let us know if there is anything else you need. Thanks.

ScottDowne:
@jkt230 question about your wording, when you say:

"we don't 1) move up or down the static elements of the page and 2) include an opt-out option."

It makes me think that we don't do 1 and 2. Is it your intent to say we don't use an opt-out option?

jkt230:
Oof sorry, good catch! I meant to say 1) DON'T the move static elements and 2) DO include the opt-out functionality.

ScottDowne:
Got it :)

ScottDowne:
@glogiotatidis ping? mind taking another look at this?

glogiotatidis:
Hey @schalkneethling what do you think about these two parts? We haven't tried using flex attrs in about:home before thus I want your input.

```css
  .flex-container {
      display: flex;
      align-items: center;
      width: 100%;
  }
 
  .flex-start {
      justify-content: flex-start;
  }
 
  .flex-space-between {
      justify-content: space-between;
  }
 
  .button-link {
      white-space: nowrap;
      margin: 5px;
      background: #D3D9DD;
      color: #000;
      font-size: 16px;
      padding: 5px 13px;
      border-radius: 6px;
  }
 
  .button-link:hover {
      background: #2B9EEF;
      color: #FFF;
  }
```

```html
     <div class="flex-container flex-start">
       {% if icon_url %}
         <a href="{{ icon_url|safe }}">
       {% endif %}
             <img class="icon" src="{{ icon }}" />
       {% if icon_url %}
         </a>
       {% endif %}
 
       <div class="flex-container flex-space-between">
         <p>{{ text|safe }}</p>
 
         {% if link_button_text %}
         <a class="button-link" href="{{ icon_url|safe }}">{{link_button_text}}</a>
         {% endif %}
       </div>
     </div>
```

These are the only differences between this template and [simple-snippet](https://github.com/mozmar/snippets/blob/master/generic-templates/simple-snippet.html)

generic-templates/button-snippet.html
```diff
@@ -0,0 +1,264 @@
...
+      {% endif %}
+            <img class="icon" src="{{ icon }}" />
+      {% if icon_url %}
+        </a>
+      {% endif %}
+
+      <div class="flex-container flex-space-between">
+        <p>{{ text|safe }}</p>
+
+        {% if link_button_text %}
```
glogiotatidis: Rename to `snippet_button_text`

generic-templates/button-snippet.html
```diff
@@ -0,0 +1,264 @@
...
+
+Variables:
+
+  @block_button_text - small text - Text to display while hovering over opt out button. Default 'Remove this'
+  @blockable: checkbox - Check to allow user to block this snippet.
+  @clickable: checkbox - Check to allow snippet to be clickable
+  @icon: image - Snippet icon. 64x64px. SVG or PNG preferred.
+  @icon_url: text - URL linked from icon and whole snippet when snippet is clickable. (optional / required when snippet is clickable)
+  @link_logo: checkbox - Check to make replacement logo clickable.
+  @link_button_text: small text - Text for a button link.
```
glogiotatidis: Add note that the button links to icon_url

glogiotatidis:
@ScottDowne this looks ok from my end. Practically you only add a button to the right of the snippet. Since this change is optional by leaving the `link_button_text` empty this should be PRed against `simple-snippet.html`. I'm trying to extend simple-snippet when possible instead of creating identical templates with minor changes so we can make all changes to the templates easier, e.g. updating the block image.

So next steps:
 - [ ] (scott) Fix minor comments
 - [ ] (scott) Wait for Schalcks input
 - [ ] (scott) PR against simple-snippet
 - [ ] (giorgos) merge 
 - [ ] (giorgos) upload to production

ScottDowne:
@glogiotatidis cool, I added the requested changes.

I agree with it being added to the simple snippet, but I wanted to push for that after you had a chance to see it coming. Glad we agree :) 

Hopefully flexbox is something we can use, if not, I can make it work without flexbox, it's just a bit more complex. Where flexbox really shines here is it allows the icon, paragraph content, and button text be dynamic sizes, and it just works.

ScottDowne:
@glogiotatidis @schalkneethling I found some time to create a backup, that didn't use flex-box, just in case. https://github.com/mozmar/snippets/pull/83/files (I made it a separate pull request, to keep this one clean)

schalkneethling:
@glogiotatidis @ScottDowne I would definitely want us to move towards using Flexbox but, depending on which versions of Fx this will actually run in, we need to do one of two additional things:

1) Add styling that will be used by the browser if it does not understand `flexbox` - i.e. styling the layout in a way that is good enough and then progressively enhance it with `flexbox`. As you mentioned @ScottDowne flex layout makes things like "variable icon width, with or without a button, and allows for the snippet to be variable length too." super easy and bullet proof.

2.) Be ok with the way the snippet will display if 'flexbox' is not supported. I would vote for option 1 myself.

According to caniuse, we have support for `flexbox` in Firefox since 47 ~ http://caniuse.com/#search=flexbox

glogiotatidis:
Since this is going to the very basic snippet template I'd like support from version 20. Is https://github.com/mozmar/snippets/pull/83 a good alternative to flexbox @schalkneethling ?

glogiotatidis:
Since we cannot use flex due to only recent browser support, i'm closing this one in favor of #83

schalkneethling:
Agreed @glogiotatidis - If we need support that for back, we need to either progressively enhance with `flexbox` or not use it at all.

ScottDowne:
I'm for not using it, let's go with #83

PR #83, opened 2016-08-26
Button snippet without flex

ScottDowne:


generic-templates/simple-snippet.html
```diff
@@ -10,6 +10,7 @@
   @icon_url: text - URL linked from icon and whole snippet when snippet is clickable. (optional / required when snippet is clickable)
   @link_logo: checkbox - Check to make replacement logo clickable.
   @replacement_logo: image - Optional. Replace logo.
+  @snippet_button_text: small text - Optional. Text for a button that links to icon_url.
```
glogiotatidis: I realized that this is more descriptive if we rename it to `button_label`. Also please change text to 
`Optional. Text for a button next to main snippet text that links to icon_url.`

generic-templates/simple-snippet.html
```diff
@@ -127,17 +146,23 @@
...
+            <img class="icon" src="{{ icon }}" />
+      {% if icon_url %}
+        </a>
+      {% endif %}
+
+      <p class="display-item">{{ text|safe }}</p>
+
+      {% if snippet_button_text %}
+      <div class="display-item">
+        <a class="button-link" href="{{ icon_url|safe }}">{{snippet_button_text}}</a>
```
glogiotatidis: Add space between  `{{`, `snippet_button_text` and `}}`

glogiotatidis:
@ScottDowne thanks for providing this non-flex version. Besides the minor nits the most important issue is that icon is not center aligned anymore.

This is a snippet with the current snippet-simple code
current-code

and a snippet with the code from this PR

new-code-with-button

When you fix this I think we can merge.

ScottDowne:
I get this: 
snippet

But I don't have the same image handy right now. I'll try a few more things, see if I can make the issue you're having.

ScottDowne:
It looks like `vertical-align: middle;` isn't cutting it, based on your screenshot.

ScottDowne:
screenshot from 2016-08-29 10-59-33

Could it be that your image has a bunch of white at the bottom from the cropping? And it is actually centered but just looks not centered? Looking at the button, it looks centered, and they use the same style rule, so I would like to get my hands on the data uri for the image you used.

glogiotatidis:
> It looks like vertical-align: middle; isn't cutting it, based on your screenshot.

Correct, that's because I didn't add a link to the icon and `vertical-align:middle` is applied to `.display-item` which in turn is only applied if there's a link. Please fix the icon alignment when there's no link and we're good to go.

ScottDowne:
Ahhh. Cool. Makes sense.

ScottDowne:
@glogiotatidis K, updated, fixed, rebased, and squashed commits.

Thanks for seeing that last one, it was a good catch.

glogiotatidis:
@ScottDowne thanks for fixing. Also made minor changes in https://github.com/mozmar/snippets/commit/e933329d431fe449fa7c25658f623892634bbd62.

glogiotatidis:
@jkt230 this is now ready for use. You have a `button_label` optional field in `Simple snippet with block and click` template.

ScottDowne:
@glogiotatidis if you're still around, I noticed there are 3 pixels different compared to the original.

They are next to the image icon, and seemingly, magical and not explicitly defined anywhere. Because of the new layout, it's more explicit, and these magic 3 pixels are gone. In the old one, the icon has a width of 53.5px and in the new one, it's an even 5px.

Is this a critical issue?

ScottDowne:
I suspect it's this that I'm seeing: https://css-tricks.com/fighting-the-space-between-inline-block-elements/

Going to dig a bit more.

But, if so, I would argue we're better off without it. Although, if we want some extra padding, we should explicitly add it.

glogiotatidis:
> Going to dig a bit more.

please check this out and let me know. currectly in serries of mtgs

ScottDowne:
It's most def the space between inline block elements. It's removed in the updates I made, which is healthier css imo, but that spacing may have been considered, so we should consider adding a few pixels left padding to the image or not.

glogiotatidis:
I'm confused. How's whitespace affecting html?

ScottDowne:
It's a tricky, not known, hard to debug gotchya withh css inline elements.

Here is a simple example to illustrate: https://jsfiddle.net/szz8or86/1/

glogiotatidis:
mind blown. 

So not all elements inside `block-snippet-overlay` where on the same line before. Why do we need to change this now?

ScottDowne:
The new layout, which is a bit more flexible, also doesn't have this little quirk.

This means, there is a small chance the initial layout was considering that space. So by "fixing it" we may be undoing a weird hack. To be safe, I suggest we add a 3px padding to the right of the icon.

glogiotatidis:
Understood. Can you please create a pull request to add the padding and
I'll review and fix tomorrow.

Thanks!
ScottDowne writes:

PR #87, opened 2016-09-09
Adding simple snippet variables for changing optional button colors.

ScottDowne:


ScottDowne:
@glogiotatidis Thoughts on this?

Adding functionality making it so the button can have it's colours configured so tests can be run as to which button colour performs better.

glogiotatidis:
I've following two thoughts on this:
 1. Too many options for the simple snippet template and there're already mistakes made with the current options we have. Unless there's a real need for this one, I'd vote against it.
 2. It's becoming super easy to create a flashy obstructive snippet which will make people rightfully upset as has happened in the past.

I'll discuss this tomorrow with Jean and come back to you.

ScottDowne:
@glogiotatidis I hear ya, and agree. I also think there is an argument that it's not really simple even with the button as is.

Bit of context: https://github.com/MozillaFoundation/Advocacy/issues/70

I think it's probably a bit of a process flaw if this is the first you're hearing about this, cause apparently this has gone by MoCo and if no ones has given you a heads up, we should call that out, because you and I have insight, and probably both need to be involved earlier.

My initial thoughts on this are:

1. Make a separate snippet with the button, that has variables for changing the button colour, and move the button out of the simple snippet. I don't see value in keeping a button in the snippet that cannot set the colour if there is a similar snippet with the button that can change colour. If a separate button snippet isn't a great idea, maybe we can consider option 2.

2. A single fundraising snippet that has the button in it, that can support variable button colours. I'm actually trying to create a set of fundraising tests based on these variables: https://github.com/MozillaFoundation/Advocacy/issues/326#issuecomment-245945353 Initially I was thinking, maybe we use two templates, and we can use the simple template for this. However, if we build a fundraising template that could do all of these, it wouldn't be hard to add all my button functionality to that.

I also think option 1 and 2 are not necessarily opposites, so we can do both if we think it's best.

glogiotatidis:
Thanks for the context Scott. I'm only responsible to make sure that things work as expected on production and I won't be calling any decisions on whether we display something or not. So my UI comments are  suggestions. On the technical front my idea is keep things as simple as possible with templates, so its easier to maintain. If the button thing is a one off we can go with (1) 

In any case any template changes must be signed off by Jean.

glogiotatidis:
@jkt230 The updated design includes the ability to change colors in the button added in #82. What do you think?

ScottDowne:
I think the open questions are:

1. To update the opt out UI/UX or not.
2. To make any changes to simple snippet.

To simplify things for now, I think we should say no on those, and build the variable colour button into the fundraising snippet, and leave simple snippet as is. Then, don't update the opt out button in either simple snippet or the fundraising snippet.

We can revisit these two questions later, but my primary concern is going to be the fundraising snippet.

jkt230:
@glogiotatidis Thanks for the flag - the button and the ability to change colors was a compromise for fundraising snippets with the MoFo team. They worked closely with Stephen Horlander (UX team) to make it not as obtrusive as done in the past so this is what they came up with. I'm fine with going forward with this as we will be testing the engagement as well as opt-out.

PR #88, opened 2016-09-23
2016 eoy fundraising snippet.

ScottDowne:
@glogiotatidis Mind taking a look at this?

### DIFF
--- a/campaigns/mofo-eoy-2016/snippet.html
+++ b/campaigns/mofo-eoy-2016/snippet.html
@@ -150,12 +150,12 @@ Variables:
     right: 5px;
     z-index: 10;
     visibility: hidden;
-    height: 24px;
-    width: 24px;
+    height: 20px;
+    width: 20px;
     cursor: pointer;
     background-image: url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAADIAAAAyCAMAAAAp4XiDAAAAeFBMVEUAAAAAAAAAAAABAQEAAAAAAAAAAAAAAAAAAAAAAADDw8OQkJAAAAAAAAAAAAAAAAD9/f38/Pz6+vr5+fnt7e3i4uLd3d3Z2dmgoKBdXV0mJibv7++amprNzc2amprv7+////9zc3OwsLD29vaPj4+xsbHCwsLBwcHrzmRyAAAAIHRSTlMABAg9HhEMNxkir38xLCkV/Pr39OHTzceOYUrliLuJ5ik+KlgAAAGhSURBVEjHnZbrdoIwDIAXWii3UuV+0wV0+v5vOEBH2ajg8v3y6PlO2pgm+TAAYD0B+HgDsBxbuXzCVbZj7VgAju16USjYhAgjz7WdrVjgKC5FHfh5kxwOSZP7QS0kVw68PJLNJQuKIy44FgGT3LbAHMKNWJviirRlkWsKNIQIqwyNZFU4BFobnihjfEFcCm92tMF83MBnfxwYjBNuchocWN6cCx938AVf5MBywxJ3KUPX0seKqnhfiavIhvlYLMM3yNjP0UDJFmcuN/zF7aI/t1I9lCFIqo1r1y+Nr+6qnXQI87iJDJZGNzra6JZOIG2Y0iWK+bt7N9JrY+Q+/1yIKWmOVy9qt3862lgGPdaeM2U4QDQ52tAEY55BhT4aHS1r/FDBeJUcTY7JwHy8jMVZgybHZGDD+KQkuHbMBiZP5YBrx2zggaCQDka5PiHJhL9yq2B6c8FslmVvKsud4u8Nxb/3xPr1E9t9yP3qIf+vXZylAkJTIrQ+QoMltHHCsCCMJMLgo49XPcTPpiF+1kOcsCrsLSSfeiGhrj305Yqwwn0Dwi/I5rQKpwMAAAAASUVORK5CYII=");
     background-repeat: no-repeat;
-    background-size: 24px 24px;
+    background-size: 20px 20px;
     background-color: transparent;
     border: 0;
   }
@@ -164,7 +164,8 @@ Variables:
     height: 100%;
     width: 100%;
     border-radius: 12px;
-    padding: 5px;
+    padding: 15px 5px 5px 5px;
+    margin-top: -6px;
     box-sizing: border-box;
   }
 
--- a/campaigns/mofo-eoy-2016/snippet.html
+++ b/campaigns/mofo-eoy-2016/snippet.html
@@ -1,0 +1,321 @@
+{#
+2016 fundraising snippet.
+
+Variables:
+
+  @text: main text - Text of snippet.
+  @icon: image - Snippet icon. 64x64px. SVG or PNG preferred.
+  @button_label: Required. Text for donate button.
+  @donation_form_url: Required. Url to the donation form.
+
+  @button_color: Default blue. Red and green also options.
+
+  @locale: Default en-US. String for the locale code.
+  @currency_code: Default usd. The code for the currency. Examle gbp, cad, usd.
+
+  @small_buttons: checkbock - Scales the buttons down.
+  @donation_amount_first: number. Required. The amount for the first button.
+  @donation_amount_second: number. Required. The amount for the second button.
+  @donation_amount_third: number. Required. The amount for the third button.
+  @donation_amount_fouth: number. Required. The amount for the fourth button.
+  @selected_button: Default donation_amount_second. Donation amount button that's selected by default.
+
+  @blockable: checkbox - Check to allow user to block this snippet.
+  @block_button_text - small text - Text to display while hovering over opt out button. Default 'Remove this'
+
+#}
+<div class="snippet" id="mofo-eoy-fundraiser-{{ snippet_id }}">
+<style>
+
+#snippetContainer div.snippet img.icon {
+    margin: 5px 8px 5px 0;
+    float: none;
+}
+
+.snippet p {
+    padding-left: 0;
+}
+
+#mofo-eoy-fundraiser-{{ snippet_id }} {
+  font-family: "Helvetica Neue", "Helvetica", "Arial", sans-serif;
+  margin-top: 6px;
+}
+
+.snippet .icon {
+    position: relative;
+    right: 0;
+    top: 0;
+    display: inline-block;
+    height: auto;
+    max-height: 64px;
+    vertical-align: middle;
+}
+
+{% if blockable %}
+  .block-snippet-overlay {
+      height: 100%;
+      width: 100%;
+  }
+{% endif %}
+
+ .snippet .display-item {
+     display: table-cell;
+     vertical-align: middle;
+ }
+ .snippet .display-container {
+     display: table;
+ }
+
+ .button-link {
+     background: #2FA0F0;
+     color: #FFFFFF;
+ }
+
+ .button-link:hover {
+     background: #A5D6F9;
+     color: #FFFFFF;
+ }
+
+{% if button_color == 'red' %}
+
+ .button-link {
+     background: #C94C4A;
+ }
+
+ .button-link:hover {
+     background: #E8B2B2;
+ }
+{% endif %}
+
+{% if button_color == 'green' %}
+
+ .button-link {
+     background: #69C9A2;
+ }
+
+ .button-link:hover {
+     background: #BEE8D7;
+ }
+{% endif %}
+
+ .top-button-link {
+     white-space: nowrap;
+     padding: 5px 13px;
+     margin: 5px 0 5px 5px;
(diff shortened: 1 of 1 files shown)


ScottDowne:
Here is the link again about what it's doing: https://github.com/MozillaFoundation/Advocacy/issues/326

There is a little table in there with some of the tests we want to run with it.

campaigns/mofo-eoy-2016/snippet.html
```diff
@@ -0,0 +1,321 @@
...
+ .button-link {
+     background: #C94C4A;
+ }
+
+ .button-link:hover {
+     background: #E8B2B2;
+ }
+{% endif %}
+
+{% if button_color == 'green' %}
```
glogiotatidis: Let's make this an `else if` to keep things tidy

glogiotatidis: then then an `else` with the default value of blue.

campaigns/mofo-eoy-2016/snippet.html
```diff
@@ -0,0 +1,321 @@
...
+          var locale = '{{ locale|default("en-US", true) }}';
+          var currencyCode = '{{ currency_code|default('usd', true) }}';
+          var btnDonate = snippet.querySelector('.button-link');
+
+          var donationURL = btnDonate.attributes['href'].value;
+          var delimiter = "?";
+          if (donationURL.indexOf('?') >= 0) {
+            delimiter = '&';
+          }
+          var dAmountFirst = '{{donation_amount_first}}';
```
glogiotatidis: Add spaces `{{ donation_amount_first }}`. same for next lines

campaigns/mofo-eoy-2016/snippet.html
```diff
@@ -0,0 +1,321 @@
...
+  @donation_amount_first: number. Required. The amount for the first button.
+  @donation_amount_second: number. Required. The amount for the second button.
+  @donation_amount_third: number. Required. The amount for the third button.
+  @donation_amount_fouth: number. Required. The amount for the fourth button.
+  @selected_button: Default donation_amount_second. Donation amount button that's selected by default.
+
+  @blockable: checkbox - Check to allow user to block this snippet.
+  @block_button_text - small text - Text to display while hovering over opt out button. Default 'Remove this'
+
+#}
```
glogiotatidis: 👍 Good docs, thanks!

campaigns/mofo-eoy-2016/snippet.html
```diff
@@ -0,0 +1,321 @@
...
+    var snippet = document.getElementById('mofo-eoy-fundraiser-{{ snippet_id }}');
+
+    snippet.addEventListener('show_snippet', function () {
+      setTimeout( function() {
+        {% if donation_amount_first %}
+
+          var btnAmounts = snippet.querySelectorAll('.donation-amount');
+          var btnSelected;
+          var amount;
+          var frequency = 'single';
```
glogiotatidis: Do we need this?

campaigns/mofo-eoy-2016/snippet.html
```diff
@@ -0,0 +1,321 @@
...
+              btnSelected.classList.remove('button-selected');
+            }
+            btnSelected = element;
+            amount = element.attributes['data-amount'].value;
+            updateLink();
+          }
+          function updateLink() {
+            btnDonate.attributes['href'].value = donationURL + delimiter + 'amount=' + amount + "&frequency=" + frequency + "&currency=" + currencyCode + '&presets=' + presets;
+          }
+          var selected = "{{ selected_button|default("donation-amount-second", true) }}";
```
glogiotatidis: Replace outer `"` with `'`

glogiotatidis: Or the other way around. You mix a lot `'` with `"`, please stick to one code style :)

campaigns/mofo-eoy-2016/snippet.html
```diff
@@ -0,0 +1,321 @@
...
+      setTimeout( function() {
+        {% if donation_amount_first %}
+
+          var btnAmounts = snippet.querySelectorAll('.donation-amount');
+          var btnSelected;
+          var amount;
+          var frequency = 'single';
+
+          var locale = '{{ locale|default("en-US", true) }}';
+          var currencyCode = '{{ currency_code|default('usd', true) }}';
```
glogiotatidis: Quotes

campaigns/mofo-eoy-2016/snippet.html
```diff
@@ -0,0 +1,321 @@
...
+            element.classList.add('button-selected');
+            if (btnSelected) {
+              btnSelected.classList.remove('button-selected');
+            }
+            btnSelected = element;
+            amount = element.attributes['data-amount'].value;
+            updateLink();
+          }
+          function updateLink() {
+            btnDonate.attributes['href'].value = donationURL + delimiter + 'amount=' + amount + "&frequency=" + frequency + "&currency=" + currencyCode + '&presets=' + presets;
```
glogiotatidis: Quotes

ScottDowne:
Thanks, working on these now.

campaigns/mofo-eoy-2016/snippet.html
ScottDowne: Fixed up. I went with double quotes for most everything, cept for any quotes inside the default calls in nunjucks, where I use single while inside a double quote string, they are like this: `"{{ currency_code|default('usd', true) }}"` this is to keep the syntax highlighters from getting confused.

campaigns/mofo-eoy-2016/snippet.html
ScottDowne: Nope, we do not. Good catch.

campaigns/mofo-eoy-2016/snippet.html
ScottDowne: Done.

campaigns/mofo-eoy-2016/snippet.html
ScottDowne: Thanks!

ScottDowne:
@glogiotatidis Fixed up. I also got the block snippet button back up. It broke because I added a unique snippet id, and not the generic one, but I didn't change the line that the button used to set visibility while hovering, so it would never show up while hovering.

ScottDowne:
Also, I figure we can squash any commits into one before we merge, but for now, I'm leaving them separate so you can see what I changed.

glogiotatidis:
> @glogiotatidis Fixed up. I also got the block snippet button back up. It broke because I added a unique snippet id, and not the generic one, but I didn't change the line that the button used to set visibility while hovering, so it would never show up while hovering.

Nice. I see that `snippet_id` is used elsewhere too. Can you please remove to keep things tidy?

campaigns/mofo-eoy-2016/snippet.html
glogiotatidis: 👍

campaigns/mofo-eoy-2016/snippet.html
```diff
@@ -0,0 +1,328 @@
...
+  @button_label: Required. Text for donate button.
+  @donation_form_url: Required. Url to the donation form.
+
+  @button_color: Default blue. Red and green also options.
+
+  @locale: Default en-US. String for the locale code.
+  @currency_code: Default usd. The code for the currency. Examle gbp, cad, usd.
+
+  @small_buttons: checkbock - Scales the buttons down.
+  @donation_amount_first: number. Required. The amount for the first button.
```
glogiotatidis: Here you say it's required and later it looks it can be skipped? https://github.com/mozmar/snippets/pull/88/files#diff-90c0ff362679dfe399e92a5a6fc3182fR217

glogiotatidis:
Contents are touching the ends of block-overlay. On simple snippet template there's more space. This is by design?

image

image

campaigns/mofo-eoy-2016/snippet.html
```diff
@@ -0,0 +1,328 @@
...
+            }
+            element.classList.add("button-selected");
+            if (btnSelected) {
+              btnSelected.classList.remove("button-selected");
+            }
+            btnSelected = element;
+            amount = element.attributes["data-amount"].value;
+            updateLink();
+          }
+          function updateLink() {
```
ScottDowne: Nope, good catch. It used to be called twice, but I simplified it a bit, and now only needed to be called once, and didn't think to just remove the function.

campaigns/mofo-eoy-2016/snippet.html
ScottDowne: Good catch here too. Fixed.

Question though, the button amounts are only optional as a group. It's all or none. I wasn't sure if I needed to explain that in the docs or not, but I can if you feel it's necessary.

ScottDowne:
I removed the snippet-id bits. I trust you that it's safe :) but I don't actually know how things changed which makes it safe, and how much I can change given there are other things I used to do to protect the snippet from leaks, like put everything inside a `div` and I'm still using a id `mofo-eoy-fundraiser` for everything. Can I update this more?

I added some padding around the block overlay, and also tweaked it a bit to remove some js, and instead used sibling selectors and hover states. This is much cleaner and removed lots of js. I also think there is room for improvement here, but for now, I'm, not going to do enhancement stuff, and may do that later after some discussion.

glogiotatidis:
> I removed the snippet-id bits. I trust you that it's safe :) but I don't actually know how things changed which makes it safe, and how much I can change given there are other things I used to do to protect the snippet from leaks, like put everything inside a div and I'm still using a id mofo-eoy-fundraiser for everything. Can I update this more?

Back in the day we use to put all snippets in the DOM of about:home and hide them all except one with `display:none`. Thus we needed to uniquely name everything to avoid breaking things. Now are packing snippets in JSON and only inserting into the DOM one snippet, thus no need to uniquely name things anymore. 

Yes you can update this as much as you want. The service doesn't care how you name things.

ScottDowne:
@glogiotatidis I couldn't figure out how to put something back into review using githubs new review flow, but this is ready again.

ScottDowne:
@glogiotatidis OK cool, so I moved the snippet out of a global `div` because it's cleaner, and not necessary. I also did a bunch of whitespace cleanup as it was getting out of hand.

I left an id on the container div that's eoy related, as I did find it slightly helpful and clear, although not needed for namespacing reasons anymore.

I did this all in a new commit, and still haven't squashed anything.

glogiotatidis:
Re: optional buttons

Let's add a checkbox "show_donation_amount_buttons"  and check against that. This will make it more clear to snippet editors (who don't read this documentation) how to configure the snippet.

ScottDowne:
@glogiotatidis good idea. I updated the docs and added a show_donation_amount_buttons

If everything checks out, feel free to rebase and squash all the commits into the first. I left separate again just so you can easily see what was changed.

Thanks!

glogiotatidis:
Close button overlays donate button. Maybe shrink the first?

image

ScottDowne:
@glogiotatidis what if I moved it to the corner, to be more like this? http://people.mozilla.org/%7Eshorlander/about-home-snippet-opt-out/about-home-test-02.html

I'm just not sure shrinking it is going to be enough.

glogiotatidis:
> @glogiotatidis what if I moved it to the corner, to be more like this? http://people.mozilla.org/%7Eshorlander/about-home-snippet-opt-out/about-home-test-02.html
> 
> I'm just not sure shrinking it is going to be enough.

I'd prefer to keep a unified experience as much as possible. I'd shrink and maybe push the button down.

ScottDowne:
k, can do.

ScottDowne:
@glogiotatidis so moving the button down is a bit of an issue, with the centering and such. I couldn't figure it out without breaking the centering in some way.

So I'm trying something a bit different, but might be better. I added some space in the overlay, to fix the button, which essentially pushed everything down, (so it stayed all centered) then I added a slight negative top margin to offset the change.

snippet

 Thoughts?

glogiotatidis:
if @jkt230 is ok with the latest screenshot, I'm ok

jkt230:
Latest screenshot looks good. Thanks!

campaigns/mofo-eoy-2016/snippet.html
```diff
@@ -0,0 +1,300 @@
...
+2016 fundraising snippet.
+
+Variables:
+
+  @text: main text - Text of snippet.
+  @icon: image - Snippet icon. 64x64px. SVG or PNG preferred.
+  @button_label: Required. Text for donate button.
+  @donation_form_url: Required. Url to the donation form.
+
+  @button_color: Default blue. Red and green also options.
```
glogiotatidis: Note: "red" is a bad option for button color because it's disruptive and also means a dangerous action. I know we default to blue so I won't block on this but please communicate this to folks asking for red buttons. 

It also made ppl upset in the past!

glogiotatidis:
Merged in 0642582a9672e75659db298396d01c35789968d2

Thanks @ScottDowne 🍻

PR #89, opened 2016-10-13
Mofo eoy 2016 QA, adding button color option and fixing spacing issues.

ScottDowne:
@glogiotatidis Hey, a few things were found/requested in the 2016 fundraising snippet.

There is a bit of spacing issues in the test without the icon, and I am missed the grey button option, which was requested.

The spacing issue is this: 
unnamed

It turns out the margin was on the icon bottom, which also isn't great in and of itself. When the icon is gone, we lost that margin, then if the text went to more than two lines, it hit the buttons. So I removed the bottom 5px margin on the icon, and instead applied it to the top of the buttons. Seems to work better.

Thoughts?

### DIFF
--- a/campaigns/mofo-eoy-2016/snippet.html
+++ b/campaigns/mofo-eoy-2016/snippet.html
@@ -8,7 +8,7 @@ Variables:
   @button_label: Required. Text for donate button.
   @donation_form_url: Required. Url to the donation form.
 
-  @button_color: Default blue. Red and green also options.
+  @button_color: Default blue. Red, green and gray are also options.
 
   @locale: Default en-US. String for the locale code.
   @currency_code: Default usd. The code for the currency. Examle gbp, cad, usd.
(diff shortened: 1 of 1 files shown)


ScottDowne:
better

For spacing, here is a screen cap with my fix, and similar snippet values.

I'm not a fan of the red button or the yellow highlight, but that wasn't my call. Hopefully those tests fail.

glogiotatidis:
> There is a bit of spacing issues in the test without the icon

Interesting I didn't expect to run this without an icon. Let's add "optional" for icon in the documentation.

glogiotatidis:
Fixed docs here d558214. 

Thanks Scott!

ScottDowne:
@glogiotatidis cool, thanks for catching the docs bit.

Question, what needs to happen to update the snippets that people see live? I tried `gSnippetsMap.clear()` and refreshed, and still had the older version. So I'm guessing something needs to happen first?

glogiotatidis:
Try clearing cache along with gSnippetsMap.clear()

ScottDowne:
Cache clearing didn't seem to do it.

Also, is 64x64 the standard icon size? It feels a little big in some cases.

PR #90, opened 2016-10-23
Selected amount button color option for gray and blue in mofo 2016 eo…

ScottDowne:
…y snippet.


ScottDowne:
@glogiotatidis review?

We'll probably want to run tests for this to see if we can prove some theories.

This one is to try blue in the selected amount colour.

Also see: https://github.com/mozmar/snippets/pull/91

PR #92, opened 2016-11-08
Adding a few new options for tests to run on 2016 eoy snippet.

ScottDowne:


ScottDowne:
@glogiotatidis hey, me again :)

I'm trying to add functionality recently requested in https://github.com/MozillaFoundation/Advocacy/issues/70#issuecomment-256108285

Comps I used as a reference:

test variant-snippet_02-01

Font sizes and another button highlight colour, nothing complicated.

The highlight is back. It was being used anyway as crufty markup in the text, and now it looks like we might need to be able to test variants, so I added it back as an explicit thing, it works better this way, so if we're going to have it, might as well work well :)

screen shot 2016-10-25 at 11 05 17 am

Also going to try a smaller icon. We also noticed if the icon is smaller than the text, the vertical centering might actually look that good. I've aligned it to the top. I didn't make this configurable, as I think it'll be better this way in all cases. If the icon is still larger or roughly the same size as the text, the text is still vertically aligned.

Thoughts?

ScottDowne:
@glogiotatidis just a friendly ping in case you missed this :)

campaigns/mofo-eoy-2016/snippet.html
```diff
@@ -24,6 +24,9 @@
   @show_monthly_checkbox: checkbox. Optional. Requires show_donation_amount_buttons. Hides or displays a monthly checkbox.
   @monthly_checkbox_label_text: Defaults to 'Make my donation monthly'. Label text for monthly chackbox.
 
+  @highlight_color: Optional. Defaults to #FFE7AC.
+  @button_font_size: Options are small and large. Small is 16px, and large is 18px. Defaults to small.
```
glogiotatidis: kinda confusing due to 'small_buttons' variable but ok

glogiotatidis:
I changed this one btw https://github.com/mozmar/snippets/commit/73ea9fb3ae6b9e8a3285d35e4786a4922b7df6c3

glogiotatidis:
Thanks @ScottDowne this is now on production! 

(and sorry for the delay, this indeed fell through the cracks)

PR #93, opened 2016-11-23
Adding updated Basket sigup snippet.

ScottDowne:
@glogiotatidis ping

Been working on this from time to time when I had free moments.

Ticket for context: https://github.com/MozillaFoundation/Advocacy/issues/588

Apology for the large ticket :(

Thoughts? Review?

🚀 🚀

### DIFF
--- a/generic-templates/newsletter-signup-snippet/snippet.html
+++ b/generic-templates/newsletter-signup-snippet/snippet.html
@@ -1,1491 +1,1468 @@
-<div class="snippet" id="signup-snippet-{{ snippet_id }}">
-  <style type="text/css">
-    @font-face {
-      font-family: 'fontello-{{ snippet_id }}';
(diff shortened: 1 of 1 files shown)


glogiotatidis:
Hey @ScottDowne i'm super busy with other stuff for the next days. What's the deadline for this?

ScottDowne:
Not sure but feels tight. it's something Jean's been trying to get out so I
would try to prioritize with her maybe.

generic-templates/newsletter-signup-snippet/snippet.html
```diff
@@ -1,1491 +1,1516 @@
...
+              <option value="LK">
+                Sri Lanka
+              </option>
 
-          <option value="LK">
-            Sri Lanka
-          </option>
+              <option value="SH">
+                St. Helena
+
```
glogiotatidis: Please change to consistent use either `'` or `"` here and in other lines

glogiotatidis: Is `translate3d` something that was recently introduced to firefox? If starting starting from which version?

generic-templates/newsletter-signup-snippet/snippet.html
```diff
@@ -1,1491 +1,1516 @@
...
+          </div>
+          <div class="transition-item">
+            <button class="back-button"></button>
+            <select class="country-input form-input" value="US" required="required">
+              <option value="" selected="selected">
+                {{ country_placeholder }}
+              </option>
+
+              <option value="US">
+                United States
```
glogiotatidis: Side note: This order is slightly tiresome if you don't leave in the UStates. I suggest that we either make the list properly sorted or we prioritize all countries that we are targeting snippet to

generic-templates/newsletter-signup-snippet/snippet.html
glogiotatidis: Same question about Fx support for this syntax with `forEach`

glogiotatidis:
Thanks @ScottDowne. Besides the couple of comments this was tested and works OK. It's although breaking the rules of the snippet design as there an input box in the page, it's also bigger than the search box and we also feature (not by default but still an option) a red button. I'll ask Jean to sign this off before proceeding.

generic-templates/newsletter-signup-snippet/snippet.html
glogiotatidis: We can change this to `https://snippets.mozilla.com/show/{{ snippet_id }}` which is actually a real URL

glogiotatidis:
Hi @jkt230. I need your sign off for this snippet since we're breaking some explicit rules as set by the Fx team:

 * Snippet features a input box and also the input box is larger than the search box
 * Snippet features a red button

ScottDowne:
I initially had it similar in size to the search button, but thought given the similar icons, it was confusing itself as the search button itself.

I can come up with a few ideas to scale it down but still not be confused with the search.

I can also just as easily make the red button green or grey (no blue because too close to search) if it makes landing this easier :) I didn't pick red.

jkt230:
@ScottDowne That would be great if you could adjust so it follows the guidelines. Thank you!

generic-templates/newsletter-signup-snippet/snippet.html
ScottDowne: Updating to " when appropriate.

ScottDowne: Yeah, been in since FF10. http://caniuse.com/#search=3d%20transforms

However, I didn't actually need the 3d version for what I was doing, so moved to 2d and now it's all the way back to version 3.5 it seems: http://caniuse.com/#feat=transforms2d

generic-templates/newsletter-signup-snippet/snippet.html
ScottDowne: Moving US into it's alphabetic spot is enough? Is that what you meant by properly sorted? If so, I'm fine with that seeing I also don't live in the US ;)

ScottDowne:
@glogiotatidis Updated. I also found a few other things I should fix. I kept the commit separate for now so you can see changes. We should squash when we commit though. Changes listed.

1. I scaled down the input so it's no longer larger or the same size as the search. I think it's a good change.
2. Green buttons.
3. Removed dead code around button_color as it's not currently used for this, and was actually just there cause I started with a snippet that had it and forgot to remove.
4. went to double quotes.
5. Moved the basket error container over a bit to align with the centered, smaller, and statically widthed container.
6. Added some right padding to the privacy policy section to avoid it ever being right up against the button.
7. Removed `[].forEach.call` no NodeList. It's not a great pattern.
8. 2d translate over 3d translate. Because unnecessary.

The transition container code is kinda verbose, I wouldn't mind going back and making it a little more concise in the future, but if it's fine for now, I say we leave it as is.

We're still breaking the one rule by having inputs in the snippet, but it's a fundamental component, and it seems is the only thing we're breaking now.

Good review, thanks.

glogiotatidis:
Awesome as always @ScottDowne , thanks for working on all these changes!

glogiotatidis:
Two last things: 

1. How about using the same arrow with the search bar for consistency?
2. How about graying out the button when the input is not selected, similar to the search bar? 

image

ScottDowne:
@glogiotatidis I initially had it closer to the search input/button, but made thought being too close to it might confuse people. Also, generally you don't want more than one primary CTA, and you want the secondary one to be different and less prominent.

It's also really easy to make these changes. So I'll make them, and you can try them tomorrow morning. I'll let you decide which of the two to keep.

jkt230:
@ScottDowne 
1) Can you confirm what the fields are? We essentially only need email, country, privacy and then subscribe. Will the country field be a list to select from?
2) Are the # of fields editable? Are the text editable?
3) How does the back end work in terms of working similarly to the email sign up landing page? Are you working with Pmac or Jdavis for the basket items? Or is this something we do within the snippet service?
4) I'm assuming if you're in the German browser, you're going to sign up and get the German version of the newsletter?

ScottDowne:
@jkt230 

1. Yeah, it's email, first name, country privacy and subscribe. First name is the only one I have that you didn't request, it seems. That was in the previous snippet so I kept it here too. Subscribe is just a button, and it's there, and privacy policy is just some text where we can put our privacy policy.
2. Number of fields are not editable. Text is editable. It's also editable so it can be localized.
3. It works just like the previous snippet. I have not worked with pmac or jdavis on this implementation specifically, but have in the past and I have a good grasp on how to send things to basket. It wouldn't be a bad idea to ping them to let them know just in case.
4. As long as we pass in locale prop like we do with the other snippet and the previous signup snippet, it should go into basket with the correct language.

jkt230:
@ScottDowne Thanks for the answers. Follow up:
1) The old sign up snippet is something I can't preview on the snippet service as I think it was removed. So is it similar to the fundraising snippet where we put in the locale field? And similarly we input the url field? 

2) I am going to assume yes but want to make sure that the opt-out or block functionality is included?

ScottDowne:
@jkt230 

1. Yes, it handles locale same as the fundraising snippet. It actually doesn't have a url field similar to fundraising because the signup happens in the snippet, no navigation away. It does have basket url prop and a newsletters prop that we sent to basket. It is using the built in sendMetric function that the snippet enables.

2. Block functionality can be enabled or disabled on the snippet. Should probably always be enabled. I'm guessing you mean opt out and block functionality is the same thing.

ScottDowne:
@jkt230 One other thing I wanted to run by you that may or may not change things.

The privacy policy bit is there, but there isn't a checkbox. By clicking singup, you agreed. It's similar to how we do it here: https://donate.mozilla.org/

ScottDowne:
@glogiotatidis ok, updated to be a bit more like search inputs.

I used one small trick where I used a gradient when I didn't do any gradient effects, because it's a good way to layer on transparent colours.

generic-templates/newsletter-signup-snippet/snippet.html
```diff
@@ -1,1491 +1,1472 @@
...
+              <option value="SD">
+                Sudan
+              </option>
 
-          <option value="PM">
-            St. Pierre and Miquelon
-          </option>
+              <option value="SR">
+                Suriname
+
```
glogiotatidis: Is this used anymore?

generic-templates/newsletter-signup-snippet/snippet.html
```diff
@@ -1,1491 +1,1472 @@
...
-    #logo-replacement-{{ snippet_id }} .play::before {
-      content: '\e801';  /* Play icon */
-      padding-left: 10px;
-    }
+{#
+Signup snippet.
+
+Variables:
+
+  @text: main text - Text of snippet.
```
glogiotatidis: Please add "use `<em>` to hightlight text" so we remember in the future

generic-templates/newsletter-signup-snippet/snippet.html
glogiotatidis: Let's remove the additional `-{{ snippet_id }}` label. We've fixed that in reporting.

glogiotatidis: That said you can remove the `sendCustomMetric` function altogether but I don't mind if you leave it there either

generic-templates/newsletter-signup-snippet/snippet.html
```diff
@@ -1,1491 +1,1472 @@
...
+  @icon: image - Optional. Snippet icon. 64x64px. SVG or PNG preferred.
+  @button_label: Required. Text for signup button.
+  @highlight_color: Optional. Defaults to #FFF8D0.
+
+  @locale: Default en-US. String for the locale code.
```
glogiotatidis: There's no default set. Either change the docs or set a default please.

generic-templates/newsletter-signup-snippet/snippet.html
ScottDowne: It's not reporting the proper line for this.

ScottDowne: I think you're talking about `callback` right?

glogiotatidis: I's targeting this line https://github.com/mozmar/snippets/pull/93/files#diff-458cd6d86e16cc8924d6fd875daffe51R1337

ScottDowne: Ah, yes, good catch.

generic-templates/newsletter-signup-snippet/snippet.html
ScottDowne: Good idea.

ScottDowne:
I did the final round of changes. I also did a rebase and squash.

glogiotatidis:
Thanks for your patience @ScottDowne :) ✌️

glogiotatidis:
Thanks for your patience @ScottDowne :) ✌️

glogiotatidis:
This is on product as "Newsletter Signup Snippet 2016" @jkt230

ScottDowne:
🚀 📬 yay thanks all.

@glogiotatidis great review as always.

@jkt230 feel free to reach out to me if you have any questions or issues setting this up.

jkt230:
Had a quick chat with both Giorgos and Scott:

1. To remove First Name field
2. To remove *Required fields (since all fields will be required)
3. Open up button color field (in description, describe the code color for red, blue and green)
4. Subscribe button will be the same color with the other buttons when highlighted 
5. Subscribe button needs to remove line break (if input "Sign up" it becomes a 2 line box)
6. Jean needs to follow up if a checkbox for privacy notice is a must have

Next steps: Jean to find out about privacy and then follow up with Scott to make all above changes.

jkt230:
@ScottDowne 

Thanks for all the changes! Two things:
1) Great news, don't need the check box for the privacy notice. I just need to change the button to say Accept instead of Sign Up.

2) I know this was the point of the slider bar but can you remove the country field for now? Or at least give me the option to keep the country field or exclude the country field?

Let me know, thanks.

ScottDowne:
Hey @jkt230 

Already landed the first round of changes: 
1. To remove First Name field
2. To remove *Required fields (since all fields will be required)
3. Open up button color field (in description, describe the code color for red, blue and green)
4. Subscribe button will be the same color with the other buttons when highlighted
5. Subscribe button needs to remove line break (if input "Sign up" it becomes a 2 line box)
6. Jean needs to follow up if a checkbox for privacy notice is a must have

I just made another request to remove country input, and pinged @glogiotatidis 

The button copy can be changed already to "accept" without changing the code, it's all done in the snippet props.

PR #94, opened 2016-11-24
Horizontal padding for longer translations

TheoChevalier:
Fixing https://github.com/MozillaFoundation/Advocacy/issues/643 for the upcoming Italian translation (and potentially other languages)

### DIFF
--- a/campaigns/mofo-eoy-2016/snippet.html
+++ b/campaigns/mofo-eoy-2016/snippet.html
@@ -119,7 +119,7 @@ Variables:
   }
 
   .bottom-button-link {
-    padding: 7px 0;
+    padding: 7px 4px;
     min-width: 122px;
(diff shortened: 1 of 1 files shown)


ScottDowne:
I'm thinking we also add a `max-width: 132px;` to ensure it never moves the button to the next line:

Example, this vs this:

good
bad

It should still fit the it locale translation in one line, this is just to protect us a bit from the unexpected, maybe?

TheoChevalier:
That’s definitely a good idea :)

ScottDowne:
Just pinging @glogiotatidis here, (who's pretty busy right now) as the person that can review this. This should be a pretty simple patch, though :)

glogiotatidis:
Thanks both!

PR #96, opened 2016-12-05
Removing country input from signup sippet.

ScottDowne:
@glogiotatidis hey, just finishing up another request from Jean from a comment here: https://github.com/mozmar/snippets/pull/93#issuecomment-264338687

Removing the county input.

Review? Understandable that the timing for this isn't great right in Hawaii. I would loop back with Jean on priorities, but this should be simple enough of a change.

Thanks!


glogiotatidis:
lgtm, thanks @ScottDowne 🌴

PR #98, opened 2016-12-14
Block EOY template after donation.

glogiotatidis:


glogiotatidis:
@ScottDowne please r?

campaigns/mofo-eoy-2016/snippet.html
```diff
@@ -293,7 +293,18 @@
     var snippet = document.getElementById("mofo-eoy-fundraiser");
 
     snippet.addEventListener("show_snippet", function () {
-      setTimeout( function() {
```
glogiotatidis: Hey @ScottDowne did we need that `setTimeout`?

campaigns/mofo-eoy-2016/snippet.html
```diff
@@ -382,8 +392,7 @@
                 elHighlighted.classList.add('active');
             }, 1000);
         }
-       });
-     });
+    })
```
ScottDowne: I think we need a semi-colon here.

campaigns/mofo-eoy-2016/snippet.html
ScottDowne: Dunno, I'm pretty sure I was doing this because older snippets were doing it.

My guess is `show_snippet` was fired too soon, and the elements we wanted to query for were not injected yet, and thus, would fail. Putting them in the `setTimeout` would allow for the next cycle.

It's worth testing.

I would feel a bit better removing it in another patch, just in case it's fixing some fringe bug from before my time?

glogiotatidis: Interestingly enough where was no timeout specified, so the code got (almost) instantly executed. I 've tested this and it works.

ScottDowne:
I think this is fine too.

Like I mentioned in the campaign based opt outs patch, I'll be one a plane for a bit starting tonight, so I won't be around for a day if anything goes wrong.

glogiotatidis:
Cool, I'll merge this but hold pushing until monday so we both have stress free weekend. Safe travels back home and thanks for your time :)

PR #101, opened 2016-12-23
Adding simple attribution prop to 2016 EOY fundraising snippet.

ScottDowne:
@glogiotatidis Adding a pretty simple attribution prop.

An example for this would be: "Mozilla fundraising uses twemoji https://github.com/twitter/twemoji to encourage donations"


glogiotatidis:
Thanks

PR #103, opened 2017-01-06
Allow enter key to progress to next step in the signup snippet.

ScottDowne:


ScottDowne:
@glogiotatidis Feature request from Jean's email.

The one part that feels weird to me, is the enter key to progress on the final step.

The final step doesn't have an imput element or any validation.

So if the enter key is hit quickly from step one and again before step two, (like a quick double click) they could submit too quickly? Might be fine though, just means they missed the policy.

It's also weird cause it makes the signup more prominent than the search. Is that ok?

Thoughts?

generic-templates/newsletter-signup-snippet/snippet.html
```diff
@@ -435,10 +435,18 @@
         var trackItemWidth = 420;
         var currentIndex = 0;
 
-        emailInput.addEventListener("focus", function() {
+        emailInput.addEventListener("click", function() {
           emailInput.classList.remove("invalid");
         });
-        function onNext(e) {
+
+        document.addEventListener("keydown", function(e) {
```
glogiotatidis: Shouldn't this get binded to `snippet` and not document to avoid having `onNext()` when user is actually using the search bar?

generic-templates/newsletter-signup-snippet/snippet.html
```diff
@@ -435,10 +435,18 @@
         var trackItemWidth = 420;
         var currentIndex = 0;
 
-        emailInput.addEventListener("focus", function() {
+        emailInput.addEventListener("click", function() {
```
glogiotatidis: Why this change?

generic-templates/newsletter-signup-snippet/snippet.html
ScottDowne: I didn't think of that, but divs don't send keydown events by default because they are not focusable by default. The document is focusable.

I can put tabindex on it to make it selectable, thus focusable, thus, fires keydown events, but feels really hacky to me.

If the search bar is focused, `onNext` shouldn't be called. So far it's the cleanest solution I could come up with, but I'm not particularly fond of it.

generic-templates/newsletter-signup-snippet/snippet.html
ScottDowne: This just clears the error state.

Before, you could only submit it by clicking, which would blur the emailInput, and if the user clicked on it again, it would reset.

With enter as submit, if it's not a valid email, it's still focused, and thus click it doesn't clear it.

You can test this like so:

1. type in an invalid email.
2. hit enter
3. click the input to fix your email

Expected: Should clear error state, because I'm fixing it.
Actual: it is still in an error state.

You can try the same steps, but instead of hitting enter, press the next button, and it should then work as expected.

glogiotatidis:
lgtm thanks @ScottDowne ! :)

PR #104, opened 2017-01-09
Hide success and error msg on back button click.

glogiotatidis:
Basket error and success messages persist if you click the back button. Fix this. 

@ScottDowne r?


ScottDowne:
I'm fine with this, but one thing I wanted to point out, to make sure it makes sense to you.

Right now on success, it hide the back button anyway. So clearing success on back is technically not needed, but having it there just in case something unexpected happens, or to cover future changes, makes sense to me.

PR #105, opened 2017-01-09
Signup snippet fixing submit bug and UX change on error.

ScottDowne:
@glogiotatidis Jean found this bug. `onNext` wasn't actually smart enough to call submit.

I'm thinking we make a check in `onNext` so if it's on the last tab, it does a submit. Thoughts?

I also noticed the UX of an error state was kinda weird. If I click the button and it triggers an error, we might as well refocus the email, and clear the email box error state when the user starts typing again. Slightly better UX.


glogiotatidis:
Thanks!

PR #106, opened 2017-03-27
[bug 1342029] Move block button more to the top and right.

glogiotatidis:


glogiotatidis:
Hey @ScottDowne could you please r?

ScottDowne:
Looking :)

ScottDowne:
Definitely an improvement R+. I think after this fix, most of the other potential improvements are around the snippet overlay highlight. I just played with a few other places for the x, but ultimately I ended up to around where you put it :)

For the overlay experience, I always liked how it's done here https://people-mozilla.org/~shorlander/about-home-snippet-opt-out/about-home-test-01.html

I would be willing to write any code to do this. That is, if you think we have a clear path towards that, from a branding/design pov?

glogiotatidis:
> For the overlay experience, I always liked how it's done here https://people-mozilla.org/~shorlander/about-home-snippet-opt-out/about-home-test-01.html
> 
> I would be willing to write any code to do this. That is, if you think we have a clear path towards that, from a branding/design pov?

I like it too! I'll merge any PRs that fix this but I'd check with :shorlander and Jean to verify that this is still wanted

glogiotatidis:
> Definitely an improvement R+. I think after this fix, most of the other potential improvements are around the snippet overlay highlight. I just played with a few other places for the x, but ultimately I ended up to around where you put it :)

Thanks!

PR #108, opened 2017-04-10
New signup snippet with test variants.

ScottDowne:
Hey @glogiotatidis 

Jean was asking if I had free time to do a new signup snippet.

Not sure yet if she has talked to you yet, so I might be ahead of the game, if so, I wouldn't worry about this until she talks to you. Otherwise, here it is :) 

Comps are here: https://mozilla.invisionapp.com/share/PVB3O3696#/screens/226586006

It's basically a new snippet, but I did copy and repurpose a lot of existing chunks of css/js from the other signup snippet and the simple snippet. It is actually quite a bit more simple than the last. There is also an added option to redirect to a url after successful signup, or use a simple success message. It also has three test variants, hover, static, and click. I think that's mostly it.

Thoughts?

### DIFF
--- a/generic-templates/newsletter-signup-snippet/snippet.html
+++ b/generic-templates/newsletter-signup-snippet/snippet.html
@@ -1,257 +1,91 @@
 {#
-Signup snippet.
+Newsletter signup snippet.
 
 Variables:
 
+  @block_button_text - small text - Text to display while hovering over opt out button. Default 'Remove this'
+  @blockable: checkbox - Check to allow user to block this snippet.
+
   @text: main text - Text of snippet. Use `em` to hightlight text.
-  @icon: image - Optional. Snippet icon. 64x64px. SVG or PNG preferred.
+  @highlight_color: Optional. Defaults to #FFE7AC.
+  @icon: image - Snippet icon. 64x64px. SVG or PNG preferred.
   @button_label: Required. Text for signup button.
-  @button_color: Optional. Defaults to #49AFFD (blue). Other suggestions #69C9A2 (green) and #C94C4A (red).
-  @highlight_color: Optional. Defaults to #FFF8D0.
-
-  @locale: Default en-US. String for the locale code.
-
   @email_placeholder: Placeholder message for email input.
   @privacy_policy: Privacy policy message.
 
   @basket_host: Default https://basket.mozilla.org. Where to send basket signups.
   @newsletters: Basket newsletter.
   @success_message: Signup success message.
+  @success_redirect: Signup success redirect url.
   @error_message: Signup error message.
 
-  @blockable: checkbox - Check to allow user to block this snippet.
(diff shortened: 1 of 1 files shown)


generic-templates/newsletter-signup-snippet/snippet.html
```diff
@@ -172,327 +198,214 @@
...
-          if (elHighlighted) {
-            elHighlighted.classList.add("active");
+
+          function changeState(status) {
+            if (status === 200) {
+              button.classList.remove("loading");
+              basketError.classList.add("hidden");
+              successMessage.classList.remove("hidden");
+              formContainer.classList.add("hidden");
+              sendMetric("signup-success");
```
glogiotatidis: Do you want to add to blocklist as well to not show the snippet to the user again?

generic-templates/newsletter-signup-snippet/snippet.html
```diff
@@ -172,327 +198,214 @@
...
+               var xbutton = xbuttons[k];
+               xbutton.addEventListener('mouseover', mouseover, false);
+               xbutton.addEventListener('mouseout', mouseout, false);
+            }
+          {% endif %}
+
+          var elHighlighted = snippet.querySelector("em");
+          // Fade in highlight on text
+          setTimeout(function() {
+            if (elHighlighted) {
```
glogiotatidis: `if` can be outside of `setTimeout`

generic-templates/newsletter-signup-snippet/snippet.html
```diff
@@ -172,327 +198,214 @@
...
+              basketError.classList.add("hidden");
+              successMessage.classList.remove("hidden");
+              formContainer.classList.add("hidden");
+              sendMetric("signup-success");
+            } else if (status === 000) {
+              //loading
+              button.classList.add("loading");
+              basketError.classList.add("hidden");
+              successMessage.classList.add("hidden");
+              {% if success_redirect %}
```
glogiotatidis: This should be in `status===200` right?

ScottDowne: Wow, yeah, goood catch.

ScottDowne:
@glogiotatidis Fixed up! :)

generic-templates/newsletter-signup-snippet/snippet.html
```diff
@@ -172,327 +199,217 @@
...
-          transitionTrack.style.transform = "translate(-" + ( currentIndex * trackItemWidth ) + "px, 0px)";
-        }
-        function onBack(e) {
-          if (currentIndex <= 0) {
-            return;
+
+          function changeState(status) {
+            if (status === 200) {
+              sendMetric("signup-success", function() {
+                addToBlockList();
```
glogiotatidis: I'll add a delay here in the case of redirect.

glogiotatidis:
Thanks!

PR #112, opened 2017-05-12
Margin and border updates to block button hover state.

ScottDowne:


ScottDowne:
@glogiotatidis Some small updates to make the simple snippet's hover state more like the new signup snippet, which I think is a bit better.

It was a pretty easy update as the snippets are pretty similar, so didn't take me long.

Thoughts? You think this is pretty objectively better, or should we ping Jean?

glogiotatidis:
This is nice. What do you think about giving the text more space on the right? (screenshot 3)

Current:

image

This PR with current width for text:

image

This PR with no width for text:

image

ScottDowne:
Yeah, I would agree with that. Making a quick update now.

Actually not sure why, but there is a default 400px width built in somewhere that's not this snippet. Maybe right in the browser itself?

ScottDowne:
Updated.

glogiotatidis:
Sweet, thanks!

PR #120, opened 2017-07-30
Simple snippet for Activity Stream.

ScottDowne:


ScottDowne:
@glogiotatidis Here is the first snippet for Activity Stream as defined in these comps.

https://mozilla.invisionapp.com/share/YD7X9Z7GX#/screens/187604366_New_Tab-Snippets

This covers both simple and with a button. It also maintains the rtl functionality.

Thanks! 🚀

activity-stream/simple-snippet.html
```diff
@@ -0,0 +1,134 @@
...
+    padding: 10px 28px;
+    border-radius: 6px;
+    box-shadow: 0 2px 0 #74777C;
+  }
+
+  .snippet .button-link:hover {
+    color: #FFF;
+  }
+
+  {% if blockable %}
```
glogiotatidis: What do you think about adding this css to the global snippets css (see https://github.com/mozmeao/snippets-service/pull/260) and bring this down to

```
  {% if blockable %}
    .snippet .block-snippet-button {
      {% if rtl %}
        left: 50px;
      {% else  %}
        right: 50px;
      {% endif %}

    }
  {% endif %}
```

This reduce the size of the payload and the complexity of the templates? Also `if blockable` maybe not required?

Thoughts?

ScottDowne:
That suggestion works for me. I think we can also include some cash on the
snippets container in there too. I'm on my way to the office and can
probably have a patch in like an hour.

ScottDowne:
@glogiotatidis ok cool, I made changes here to account for the css now in https://github.com/mozmeao/snippets-service/pull/261

glogiotatidis:
Great work as always @ScottDowne . Pushing this to production 👍

PR #128, opened 2017-12-05
Activity stream simple snippet height changes for eoy fundraising.

ScottDowne:
@glogiotatidis Tall simple snippet from https://mozilla.invisionapp.com/share/PDEPWO1SK#/screens/267006747

I decided to put these changes in the existing snippet instead of trying to fit them in a fundraising snippet that worked like the simple snippet or with buttons.

It ups the height as an option. 

I increased the x button with css overrides and added a bit more padding beside the button and icon to better match the new size.

I also fixed https://github.com/mozmeao/snippets/issues/127

I also made the button colour configurable, and supplied defaults.

The defaults better match what is now in http://design.firefox.com/photon/components/buttons.html after chatting with Aaron.

Thoughts? The fundraising with amount buttons is happening in another patch.

### DIFF
--- a/activity-stream/simple-snippet.html
+++ b/activity-stream/simple-snippet.html
@@ -8,16 +8,22 @@ Variables:
   @icon: image, optional - Snippet icon. 64x64px. SVG or PNG preferred.
   @button_label: text, optional - Text for a button next to main snippet text that links to button_url. Requires button_url.
   @button_url: text, optional - A url, button_label links to this.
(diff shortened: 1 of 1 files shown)


ScottDowne:
Yup that's intended. It otherwise looked pretty small in the much bigger
space. I tried to match the new comps.

glogiotatidis:
Got it, thanks!

glogiotatidis:
pushed to production with the name "eoy 2017"

glogiotatidis:
push to production

PR #129, opened 2017-12-05
Adding 2017 eoy fundraising snippet for Activity Stream.

ScottDowne:


ScottDowne:
Hey @glogiotatidis here is the fundraising snippet with buttons based on the comps in here: https://mozilla.invisionapp.com/share/PDEPWO1SK#/screens/267006747

It's mostly a port from the 2016 fundraising snippet, so buttons should more or less work the same.

The only weird thing is the `setTimeout` this is due to some values added via the snippet service to the link after the show_snippet events in here. By using a setTimeout it delays getting the url so when we try to get it, the proper query params are there.

Thoughts?

glogiotatidis:
Block button is larger than typical block button, is this intended?

glogiotatidis:
> Block button is larger than typical block button, is this intended?

answer in https://github.com/mozmeao/snippets/pull/128#issuecomment-349228530

glogiotatidis:
Thanks!

PR #130, opened 2017-12-07
Adding newsletter subscribe snippet for AS.

jpetto:
This PR adds a snippet that allows users to subscribe to a given newsletter.

There are 4 different scenes/views:

1. Initial - intro text with button that reveals scene 2
2. Subscribe form
3. Subscribe success message
4. Subscribe failure message

One potentially tricky part of this snippet is making sure the configured basket newsletter id is valid. It would be great to provide a pre-populated drop-down list in the admin view, but I don't think that's possible. Maybe we can provide a link in the admin to accepted/existing basket ids?

### DIFF
--- a/activity-stream/newsletter-subscribe.html
+++ b/activity-stream/newsletter-subscribe.html
@@ -1,0 +1,364 @@
+{#
+Activity Stream snippet for email signup.
+
+Variables:
+
+  @blockable: checkbox - Check to allow user to block this snippet.
+  
+  @scene1_icon: image, optional - Snippet icon. 64x64px. SVG or PNG preferred.
+  @scene1_button_label: text - Text for the button next to main text that opens the email form.
+  @scene1_button_background_color: hash, optional - Button background color. Default #D7D7DB.
+  @scene1_button_color: hash, optional - Button text color. Default #0C0C0D.
+  @scene1_text: text - Text for scene1/initial view of the snippet.
+  @scene1_title: text, optional - Title displayed before scene1_text.
+  @scene1_title_icon: image, optional - Title icon. 16x16px. SVG or PNG preferred. Grayscale.
+  
+  @scene2_text: text - Text for scene2/form view.
+  @scene2_title: text, optional - Title displayed before scene2_text.
+  @scene2_newsletter: text - Newsletter/basket id user is subscribing to. Must be a value from the 'Slug' column here: https://basket.mozilla.org/news/. Default 'mozilla-foundation'.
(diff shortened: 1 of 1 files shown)


jpetto:
@glogiotatidis Let me know what you think here, both in terms of the actual code and the basket id issue. Thanks!

glogiotatidis:
> One potentially tricky part of this snippet is making sure the configured basket newsletter id is valid. It would be great to provide a pre-populated drop-down list in the admin view, but I don't think that's possible. Maybe we can provide a link in the admin to accepted/existing basket ids?

That is a valid point. The snippets team has been managing succefully those free text fields for quite a while so I think they will be fine.  I make the description of each variable available to them, in the admin area, so if you have a link or list with the accepted basket ids please update [this line](https://github.com/mozmeao/snippets/pull/130/files#diff-00a6d612fc881bc7ac52c6d96cf81772R18)

Those said I'd love to get more field types in the future... when I find the time.

glogiotatidis:
This is good work and I think we're close. I do have some visual concerns:

 - It seems you're defining a larger font size than the default. We should try and stick to the defaults to have snippet uniformity to the maximum extend. This applies for other properties that you may have altered from the default.

 - The block snippet button in the second scene is different from the typical 'x' on the right side. Have you checked with Aaron this design? Do you have mock-up on which you based your implementation on?

jpetto:
I am following the design spec provided in [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1404542#c11), which specifies 14px fonts and the alternate dismiss button on scene 2. The Sketch file was posted by Aaron Benson - is that the same Aaron you're referring to?

jpetto:
Here's a screen cap of the bit of the Sketch file detailing this form:

screen shot 2017-12-08 at 9 30 04 am

jpetto:
Added a link to the available newsletter ids.

glogiotatidis:
> I am following the design spec provided in the bug, which specifies 14px fonts and the alternate dismiss button on scene 2. The Sketch file was posted by Aaron Benson - is that the same Aaron you're referring to?

Yes and in this case I'm fine with it.

glogiotatidis:
@jpetto one last thing we need to change before we merge:

We globally changed the blocking button to only appear on hover in https://github.com/mozmeao/snippets-service/commit/ac832de08ba19577b42999aa42f46f47236213cd

which breaks stage 2 block button. Can you please update?

jpetto:
Updated to always show scene 2 block button. Thanks @glogiotatidis!

glogiotatidis:
Thanks @jpetto 🎄

PR #131, opened 2017-12-15
Updates to 2017 eoy activity stream snippet for height and size changes.

ScottDowne:


ScottDowne:
Hey @glogiotatidis 

There was some chat around running some size related tests on the snippet. Comps are here: https://mozilla.invisionapp.com/share/9UEZNE7AW#/screens/269619620_New_Tab_-_Donation_Snippet_XXL

I decided to add a test string in the yaml for this, instead of making each little bit configurable for simplicity sake. It's much easier to setup a snippet this way. I'm guessing once we know a bit more, we can decide to make some of these changes smaller variables if we want.

I also added a bit of padding bottom to the main content behind the snippet, because the increased height meant some content in activity stream was stuck behind the snippet.

thoughts?

glogiotatidis:
@ScottDowne icon size specs for the variations are the same?

PR #132, opened 2017-12-19
Add text color variable to mofo eoy fundraising snippet.

ScottDowne:


ScottDowne:
@glogiotatidis Forgot this before, but in order to have darker colours like in the comps, we need to be able to modify the text colour, and also have the x button match. For now we're just running light grey tests.

It was a bit tricky getting the x button to have configurable colour to match, as it is an svg image. I explored a few options but found a pretty simple way using just css. Thoughts?

glogiotatidis:
in prod

PR #133, opened 2018-01-08
[fix bug 1404542] Adds a send to device snippet for AS.

jpetto:
@glogiotatidis - Along with general review, there's a question for you (commented in code) regarding the availability/use of `{{ snippet_id }}`.

### DIFF
--- a/activity-stream/send-to-device.html
+++ b/activity-stream/send-to-device.html
@@ -1,0 +1,488 @@
+{#
+Activity Stream snippet for send to device mobile download.
+
+Variables:
+
+  @blockable: checkbox - Check to allow user to block this snippet.
+
+  @include_sms: checkbox - Defines whether SMS is available.
+  @locale: text - String for the locale code. Default 'en-US'.
+
+  @message_id_sms: text - Newsletter/basket id representing the SMS message to be sent.
+  @message_id_email: text - Newsletter/basket id representing the email message to be sent. Must be a value from the 'Slug' column here: https://basket.mozilla.org/news/.
+  
+  @scene1_button_background_color: hash, optional - Button background color. Default #D7D7DB.
+  @scene1_button_color: hash, optional - Button text color. Default #0C0C0D.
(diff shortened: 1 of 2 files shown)


activity-stream/send-to-device.html
```diff
@@ -0,0 +1,497 @@
...
+  
+        function sendEmailToBasket() {
+          const formData = new FormData();
+          formData.append('email', input.value);
+          formData.append('newsletters', '{{ message_id_email }}');
+          // REVIEWER (Giorgos, most likely :)
+          // In a previous send to device snippet (/campaings/mobile/mobile-snipped-widget.html),
+          // the URL below has "{{ snippet_id }}" appended to the end. Is this value still available?
+          // It appears to be empty testing locally in simple_template_server, but maybe it's available
+          // in the real application?
```
glogiotatidis: `{{ sinppet_id }}` works on both simple_template_server and the real application, just tested it. Note that `snippet_id` on the dev servers starts with `0` so maybe you mistook it for url encoding?

activity-stream/send-to-device.html
```diff
@@ -0,0 +1,497 @@
...
+          formData.append('source_url', encodeURIComponent('https://snippets.mozilla.com/'));
+  
+          // Maybe change to the following?
+          // formData.append('source_url', encodeURIComponent('https://snippets.mozilla.com/{{ snippet_id }}'));
```
glogiotatidis: I suggest `https://snippets.mozilla.com/show/{{ snippet_id }}` which displays a preview of the snippet. (if snippet is public)

activity-stream/send-to-device.html
jpetto: Totally lost it in the URL encoding. 😁

jpetto:
@glogiotatidis Updated. Thanks!

glogiotatidis:
Thanks!

glogiotatidis:
Both pushed to production

PR #137, opened 2018-01-12
[fix #135] Add locale string to newsletter AS snippet.

### LINKED ISSUE #135: Need locale input field in: Newsletter Template for Activity Stream
Hi @jpetto noticed that we can't input locale which we need to make sure the right newsletter is being sent.

jpetto:


glogiotatidis:
Thanks!

PR #138, opened 2018-02-02
[fix bug 1423414] Add FxA activity stream snippet.

jpetto:


glogiotatidis:
This is good work @jpetto. By re-reading the discussion in the bug and since now we're directly linking to accounts.firefox.com I'm wondering why -in the spirit of code reuse and less templates to maintain- we can't re-use the simple snippet template and let the Snippets Team use a link like `http://accounts.firefox.com/?action=email&service=sync...`?

jpetto:
> This is good work @jpetto. By re-reading the discussion in the bug and since now we're directly linking to accounts.firefox.com I'm wondering why -in the spirit of code reuse and less templates to maintain- we can't re-use the simple snippet template and let the Snippets Team use a link like http://accounts.firefox.com/?action=email&service=sync...?

I think we do need a separate template here, primarily for the user to enter their email address. We also have some JS that parses the user's user-agent for their Fx version and sticks that in the form. This template is very similar visually to the newsletter snippet, but I think has just enough customization to warrant another template.

jkt230:
@jpetto Question for you, if you are not signed in and go to http://accounts.firefox.com, will the email address be prepopulated since you've already entered it in the snippet? [can we make it so it's that way?]

jpetto:
> Question for you, if you are not signed in and go to http://accounts.firefox.com, will the email address be prepopulated since you've already entered it in the snippet? [can we make it so it's that way?]

Yes, the email address entered into the snippet will be carried over to whichever form is displayed on `accounts.firefox.com` - sign-up or sign-in.

glogiotatidis:
> I think we do need a separate template here, primarily for the user to enter their email address (which is used by accounts.firefox.com to display either the sign up or sign in view). We also have some JS that parses the user's user-agent for their Fx version and sticks that in the form. This template is very similar visually to the newsletter snippet, but I think has just enough customization to warrant another template.

Makes perfect sense, thanks. 

Please update with the new metrics from #140 and i'll review

jpetto:
Metrics added!

jpetto:
Updated "Nevermind" button to "Dismiss" and added `data-metric="snippet-scene2-blocked"`.

glogiotatidis:
@jpetto this looks slightly different from the send to device template (input is more centered) and the `send` button is always enabled (which i believe is fine imho, we can leave email checking to fxaccounts service). 

I'll merge and push this one to unblock fxa related snippets. If we decide we want those two templates visually closer feel free to file a new pr.

Thanks! 🍰

PR #142, opened 2018-02-14
Update send to device snippet to allow long DE phone numbers.

jpetto:
Also sends `locale` and `country` to basket for SMS message translation.

Fixes mozmeao/snippets-service#354.


glogiotatidis:
Thanks!

PR #144, opened 2018-04-24
Fix Bug 1449653 - use theme variables from activity stream

rlr:
r? @k88hudson and/or @glogiotatidis ?

### DIFF
--- a/activity-stream/fxa.html
+++ b/activity-stream/fxa.html
@@ -104,7 +104,7 @@ Variables:
   .snippet .button-link {
     background: {{ scene1_button_background_color|default("var(--newtab-button-secondary-color, #D7D7DB)", true) }};
     border-radius: 2px;
-    border: 0;
+    border: 1px solid var(--newtab-border-primary-color, transparent);
     color: {{ scene1_button_color|default("var(--newtab-text-primary-color, #0C0C0D)", true) }};
     font-size: 14px;
     font-weight: 600;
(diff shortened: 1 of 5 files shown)


rlr:
I should probably update the default colors below to use our bottom variables?


```
  .snippet .button-link {
    background: {{ button_background_color|default("#D7D7DB", true) }};
    color: {{ button_color|default("#0C0C0D", true) }};
```

k88hudson:
Agreed, button colors would be good :+1:. You also probably want to take a look at other templates in https://github.com/mozmeao/snippets/tree/master/activity-stream, although some might not need to be converted (End of Year for example) @glogiotatidis what do you think?

rlr:
oh derp. It is supposed to have a border. @glogiotatidis How do I test that specific snippet locally?

glogiotatidis:
> oh derp. It is supposed to have a border. @glogiotatidis How do I test that specific snippet locally?

You can try https://github.com/mozmeao/snippets-service/tree/master/simple_template_server

glogiotatidis:
@rlr is this ready for review or you want to test first?

rlr:
@glogiotatidis I *think* it's good but I want to try to test it first for sanity.

rlr:
OK @glogiotatidis, thanks. Tested out and buttons look OK now:

screen shot 2018-05-08 at 12 14 26 pm

glogiotatidis:
Merged here with some fixes on the second stage buttons.
https://github.com/mozmeao/snippets/commit/235b00cb7709a236b0eb80998af6a0214f9b27d3

PR #146, opened 2018-07-23
Use inset-* alongside offset-* properties to support both nightly and release channels

piatra:
Follow up to https://github.com/mozmeao/snippets-service/pull/610 not sure if these are still in use but just to be sure.


glogiotatidis:
Thanks! :cake:

PR #148, opened 2018-08-27
[fix #661] Add Firefox Rocket send to device AS template.

jpetto:


alexgibson:
Some link color contrast issues:

screen shot 2018-08-29 at 09 29 43

Also on the privacy text:

screen shot 2018-08-29 at 09 30 01

alexgibson:
When I hit submit, I was a bit confused if the form had been successful or not. The button changes to "Email Sent" temporarily, but while this happens the button is in the disabled state and a bit hard to read. It also disappears quite quickly, and the email field is left with a red border, a bit like an error occurred?

screen shot 2018-08-29 at 09 30 44

alexgibson:
In the default state the form area looks pretty grayed out, almost as if it's disabled.

screen shot 2018-08-29 at 09 34 52

Could this convert better if we made the field a different background color and the button a bit brighter perhaps?

screen shot 2018-08-29 at 09 36 45

alexgibson:
Error text also has color contrast issues:

screen shot 2018-08-29 at 09 45 32

Of course, now I realize that these issues are likely due to the fact I have the Firefox dark-theme enabled, which Activity Stream now supports 🤔

activity-stream/send-to-device-rocket.html
```diff
@@ -0,0 +1,509 @@
...
+        sendMetric('activation-email-success');
+
+        formSubmit.textContent = '\u2713 Email sent!';
+
+        resetButtonText();
+      }
+
+      function performFetch(fetchRequest) {
+        formSubmit.dataset.originalText = formSubmit.textContent;
+        formSubmit.textContent = 'Sending...';
```
alexgibson: Are hard-coded English strings an issue here?

activity-stream/send-to-device-rocket.html
```diff
@@ -0,0 +1,509 @@
...
+            <a href="{{ google_play_link_url }}" id="google-play-link">{{ google_play_link_text|default('visit the Play Store', true) }}</a>
+          </p>
+        </section>
+      </div><!--/.section-content-->
+    </section>
+
+    <section id="send-to" class="section">
+      <div class="section-content">
+        <form action="" id="send-to-device-form">
+          <input id="email" name="email" type="text" required="required" placeholder="{{ form_input_placeholder|default('Enter your email address...', true) }}" />
```
alexgibson: From an a11y perspective, it would be nice to have a label associated with the email field here (even if hidden visually?)

activity-stream/send-to-device-rocket.html
```diff
@@ -0,0 +1,509 @@
...
+          };
+          var callNow = immediate && !timeout;
+          clearTimeout(timeout);
+          timeout = setTimeout(later, wait);
+          if (callNow) func.apply(context, args);
+        };
+      };
+
+      // make sure #snippet-container is positioned so the form is hidden just below the bottom of the window
+      function repositionSnippetContainer(specifiedOffset) {
```
alexgibson: What you have here totally works so may not be worth changing, but could you avoid all the resize awareness logic if you toggled a hidden div within the snippet container, and then animated it's `max-height` using a CSS keyframe animation? 

Disclaimer: I have no idea how snippet sizing is calculated and have not tested this at all, but if the browser can do the work here it could save some bytes?

activity-stream/send-to-device-rocket.html
```diff
@@ -0,0 +1,509 @@
...
+  }
+
+  @media (min-width: 766px) {
+    .snippet img {
+      display: block;
+    }
+  }
+
+  .snippet section a {
+    text-decoration: underline;
```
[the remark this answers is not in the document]
jpetto: Unfortunately, not really. AFAIK, most snippets (including this one) are simply modified versions of the most similar existing snippet. We definitely _should_ have conventions. Time and priorities though...

activity-stream/send-to-device-rocket.html
```diff
@@ -0,0 +1,509 @@
...
+        if (newOffset !== undefined) {
+          // this updates a CSS variable which impacts the transform: translateY on #snippets-container
+          style.setProperty('--yOffset', newOffset + 'px');
+        }
+      }
+
+      // check if it is a valid email
+      function checkInput(val) {
+        // http://emailregex.com/
+        const email_re = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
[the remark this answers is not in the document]
jpetto: This was one of those carry-over copy pasta jobs. I simplified to the one we use on bedrock. Good call.

activity-stream/send-to-device-rocket.html
jpetto: Good catch. Made these strings configurable.

activity-stream/send-to-device-rocket.html
jpetto: Good call. Done.

activity-stream/send-to-device-rocket.html
jpetto: This is a good approach to try, but I'm going to pass for now due to time constraints.

jpetto:
The dark theme issues are potentially very real and are larger than just this snippet. I'll file an issue to look at this, but I don't think it should block here.

PR #150, opened 2018-08-30
[fix #661] Update send to device card template (part 2):

jpetto:
- Fix Play Store link color
- Small default string changes for error message and privacy link
- Remove 'sending' button state
- Re-work 'sent' button state


jpetto:
This version is already in production and has been tested/verified by Michelle A., so we should just need a bug/sanity check here.

PR #151, opened 2018-08-31
Fix #661: Remove google play link from send-to-device-card

pmac:


pmac:
These changes are now live in the snippets service. Merging.

PR #152, opened 2018-10-18
Fix Bug 1488314 - Add scope to snippet CSS to prevent changing Activi…

piatra:
…ty Stream layout

The rule also affected AS cards. 
bildschirmfoto am 2018-10-18 um 20 08 40


glogiotatidis:
Hey @piatra you've commited your vim's `swp` file. Remove so we can proceed. Thanks!

glogiotatidis:
Thanks @piatra.

Pushed to prod

PR #153, opened 2018-10-29
Automatically block snippet after donate button gets clicked.

glogiotatidis:
Issue https://github.com/mozmeao/snippets-service/issues/755

### DIFF
--- a/activity-stream/mofo-eoy-2017.html
+++ b/activity-stream/mofo-eoy-2017.html
@@ -17,7 +17,7 @@ Variables:
   @background_color: hash, optional - Snippet background color. Default #FFFFFF.
   @text_color: hash, optional - Text color. Default #000000.
   @highlight_color: hash, optional - Paragraph em highlight color. Default #FFE900.
-  
+
   @locale: Default en-US. String for the locale code.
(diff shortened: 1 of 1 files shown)


activity-stream/mofo-eoy-2017.html
```diff
@@ -278,6 +278,16 @@
...
 
+      btnDonate.addEventListener("click", function(event) {
+          if (event.target.dataset.blocked !== 'true') {
+              event.preventDefault();
+              event.stopPropagation();
+              event.target.dataset.blocked = 'true'
+              addToBlockList();
+          }
+          event.target.click();
+      });
```
jpetto: If the intent is to call `addToBlockList()` before the page navigates elsewhere, I think you could do this a little simpler:

```
event.preventDefault();
event.stopPropagation();
addToBlockList();

// give addToBlockList 250ms to do it's thing before navigating away
setTimeout(function() {
    document.location = event.target.href;
}, 250);
```

I don't think we need any `data-*` attributes set here, as the page will navigate away anyway, thereby wiping any dynamically created `data-*` attributs.

jpetto: Ah, I see. Makes sense. Carry on!

